PR #32, opened 2022-12-07
Ry/CI_setup

Ryno95:
Added following files (Also the only file to checkas the rest are imports and other garbage):
- .github/workflows/ft_transcendence.yml

This contains a workflow that will:
- Start a linux container
- Check out our code
- Install dependencies and run backend unit tests
- Install dependencies and run frontend unit tests

This will happen every time a branch is pushed to.  This will eventually be added to the necessary checks to be able to merge the branch into the main.

Below at the status check you will see:
"""All checks have passed"""
You can them click on show all checks to see the tests that have run or by going to the ""Actions"" tab.

They are split up as two different jobs to save time as different jobs can run in parallel

PR #34, opened 2022-12-13
Vi/typeorm db connection

### LINKED ISSUE #33: NestJS can not connect to database
The containers are all up and running but there is no database connection yet.

Needs:
- [x] typeorm installation
- [x] connection to database
- [x] fix unit tests
- [ ] fix e2e tests

vincentvis:
This should be able to make everyone connect to the database.

- Installs typeorm and postgresql packages (pg package, not postgresql package!!)
- Fixes backend access to .env file
- uses npm ci instead of npm install in the container (forces the usage of the lockfile)

In addition to this I changed the prettier config to use tabs instead of spaces and changed this for the first few relevant files (replaced spaces with tabs) so quite a few of the changes are just that, spaces -> tabs.

(should be able to merge, without conflicts with Ry/CI_setup #32 branch)

closes #33


backend/.prettierrc
```diff
@@ -1,4 +1,5 @@
 {
   "singleQuote": true,
-  "trailingComma": "all"
-}
\ No newline at end of file
+  "trailingComma": "all",
+  "useTabs": true
```
Ryno95, line 4: Aaaah zo doe jij dat! Sexy hoor! :)

compose.yml
```diff
@@ -22,17 +22,22 @@ services:
...
       - ./backend/shared:/src/app/shared:ro
     restart: unless-stopped
+    env_file:
+      - .env
     depends_on:
       - database
 
   database:
     container_name: database
+    command: postgres -c stats_temp_directory=/tmp
```
Ryno95, line 32: I tried searching for what this does, couldn't anything within 5 minutes, could you give a short explanation on this?

Ryno95:
image

Tests are failing cause the import is not the same as the class name as defined in typeorm.service.ts.

Also did you run this test locally before pushing as it doesn't run as far as I can see...?

compose.yml
vincentvis: postgres stores so called temporary statistics data [^1][^2]
we are probably never going to use it, but with the default settings postgres did not have access permissions to the default location it tries to write these stats to. So this solves that (also kinda mentioned in a [commit message](https://github.com/Liryvian/ft_transcendence/pull/34/commits/dfaa2cefbb88e22ee76f7a683c52c5ce74ad84c4) but i 100% forgive you for not finding that..)

--
sources:
[^1]: https://postgresqlco.nf/doc/en/param/stats_temp_directory/
[^2]: https://www.postgresql.org/docs/14/runtime-config-statistics.html#GUC-STATS-TEMP-DIRECTORY

vincentvis:
@Ryno95 
I did not run the tests locally at all!
but I will do that now and learn something new

compose.yml
Ryno95: Aaaaa okay yes thanks for elaborating!

Ryno95 (approved):
Looks good Vincent!
Nice fixes for the env files :)

livlaming (approved):
voor zo ver ik kan zien prima de bima, het is allemaal lekker nieuw.

PR #46, opened 2022-12-20
added abstract class

Ryno95:
Split-off a small pr for abstract.service class.

Contains the basic CRUD also as created when generating a resource with the Nest CLI.

No tests as it needs a implementation of the abstract class to run tests.


vincentvis (approved):
👍

vincentvis (changes requested):
filename!
can you change the filename to `abstract.service.ts` ?

backend/.gitignore
```diff
@@ -1,6 +1,7 @@
 # compiled output
 /dist
 /node_modules
+package-lock.json
```
vincentvis, line 4: the package-lock.json file is used by `npm ci` to install the package with the versions specified in the lockfile. Without a lockfile it doesn't work. Is there a specific reason you would not want the package-lock file in the repo?

Also, gitignore rules only affect untracked files. Adding an already tracked file to gitignore does not work, you'll have to untrack the file in question for the gitignore rule to take effect.

vincentvis (changes requested):

vincentvis (approved):

PR #47, opened 2022-12-20
Ry/test examples

### LINKED ISSUE #35: Back-End Test setup

Ryno95:
Ok so the testing setup seems to work :) 

We have a in memory db for testing that we can also seed with the same data every time, which is super nice! 

So I've added a folder `test_example`. In here you will find a nest generated resource `Animal`. An Animal entity contains only a NAME and an ID to keep things simple. 

I've also extended the animal.service with the abstract.service that Anna started setting up. 

It contains one basic test for each controller, and one test for the service that is unique to the animal.service file. In think continuing we only have to test controllers/services that are not in the abstract class.

The e2e tests contain one basic test for each endpoint of the animal class. As the content is tested in the  unit tests, we wanna keep the e2e tests nice and simple, so check that you are getting the right HttpStatusCode and maybe that the response contains a body. It is also easiest to check for requests that should go wrong, in the e2e tests. As simple as checking httpStatusCodes. (4xx, 5xx)

Please refer to the [wiki](https://github.com/Liryvian/ft_transcendence/wiki/Back-end-testing-setup#setting-up-back-end-unit-and-e2e-tests) for a complete tutorial on setting up the back-end unit and e2e tests!

Closes #35


backend/src/test_example/animal.service.ts
```diff
@@ -0,0 +1,22 @@
+import { HttpException, HttpStatus, Injectable } from '@nestjs/common';
+import { InjectRepository } from '@nestjs/typeorm';
+import { AbstractService } from '../shared/abstrct.service';
```
vincentvis, line 3: filename :)

backend/test/animal.e2e-spec.ts
```diff
@@ -0,0 +1,108 @@
...
+                    .expect(HttpStatus.OK)
+                    .expect((response: request.Response) => {
+                        const {
+                            affected
+                        } = response.body;
+                        expect(affected).toEqual(1);
+                    })
+                })
+            })
+});
```
vincentvis, line 108: so far this seems to be only testing behaviour that is going well.
should we add tests for things that don't go well?

GET/DELETE/PATCH for ID's that don't exist
POST with wrong/empty body

something like this:
```ts
describe('/test/6 (GET)', () => {
    it('should return a 404', () => {
        return request(app.getHttpServer())
            .get('/test/6')
            .expect(HttpStatus.NOT_FOUND);
    });
});
```

now is a failing test, since the service just returns NULL, which is what is also returned to the enduser.. So we get a 200 OK response with an empty body.. 
Not sure if that is the behaviour we want.. Also not sure if we want to respond with 404.. but still

vincentvis (commented):
i made some changes in this branch /vi/rytestExamples

also, did you install prettierjs? (and enable format on save?)
since my prettierjs changes everything :P

backend/src/shared/abstract.service.ts
```diff
@@ -1,27 +1,35 @@
...
 	}
 
-	async create(data): Promise<any> {
+	async create(data): Promise<T> {
 		return this.repository.save(data);
 	}
 
-	async findOne(condition): Promise<any> {
-		return this.repository.findOne(condition);
+	async findOne(condition) {
```
vincentvis, line 16: it seems like this can still have the return type, or is there a reason it isn't on this method?

```ts
async create(data): Promise<T> {
```

vincentvis (approved):

backend/src/typeorm/typeorm.service.ts
```diff
@@ -28,7 +31,7 @@ export class TypeOrmConfigService implements TypeOrmOptionsFactory {
 			synchronize: true, // never use TRUE in production!
```
livlaming, line 31: so we need to take this out in the end?

livlaming (approved):
looks good! I do need to look into the testing etc, and so my check is unfortunately not as thorough as Id prefer, but I think for now it's nice to merge this asap. 
Looking forward to work with the abstract class and from there get into the testing! <3

Ryno95 (commented):
Looks good! Thanks for the clear improvements V!

Issue #38, opened 2022-12-19
Create game_invites table with basic CRUD endpoints


Ryno95:
Create the crud resource for game invites (With commented fk)

PR #50, opened 2023-01-05
Ry/game invites

Ryno95:
Basic entry points for /game-invites all using the abstract class and nothing from the db is linked, tried to keep it as simple as possible.

Files created == nest g resource games.

Added a unit test file and an e2e test file. With the db connection and a simple test in both cases. Also added unit test for the DTO. Very similar to currentGames tests, but I don't trust the Dto's yet ehe!

Also removed the update controller and UpdateDto as I imagine it'll be something const until removed or accepted.

#38


vincentvis (approved):
it indeed looks a lot like the other PR we've just looked at :)

backend/src/pong/game_invite/entities/game-invite.entity.ts
```diff
@@ -0,0 +1,16 @@
+import { Check, Column, Entity, PrimaryGeneratedColumn } from 'typeorm';
+
+@Entity('game-invites')
```
livlaming, line 3: minor detail but for the constancy I thought entity names in snake case?

(copy pasta from slack)
case:
Class names = PascalCase
File names = kebab-case  (Doet nest ook bij het generaten van resources)
Column names + Table names =  snake_case (postgres bewaard alles in lower case) (edited)

backend/src/pong/game_invite/game-invite.controller.ts
```diff
@@ -0,0 +1,36 @@
...
+	@Get()
+	findAll() {
+		return this.gameInvitesService.findAll();
+	}
+
+	@Get(':id')
+	findOne(@Param('id') id: number) {
+		return this.gameInvitesService.findOne({ where: { id } });
+	}
+
```
livlaming, line 31: probably on purpose, but is it correct that there is no update functionality? 
as in we can't update game-invite? doe sort of makes sense(?)

backend/src/pong/game_invite/entities/game-invite.entity.ts
Ryno95: Yes!

backend/src/pong/game_invite/game-invite.controller.ts
Ryno95: Exactly, there isn't any info to update as the users stay constant, until the invite is removed :)

PR #51, opened 2023-01-06
Vi/feat abstract class all relations

vincentvis:
Some improvement suggestions to the abstractService.

- allow relationships to be set to the findAll method (optional)
- allow creation and deletion through an array (create/delete many of T in one go)

Added tests for this through the Animal and changed the Animal to reflect those changes.

### DIFF
--- a/backend/src/test_example/animal.controller.ts
+++ b/backend/src/test_example/animal.controller.ts
@@ -8,6 +8,7 @@ import {
 	Delete,
 	HttpException,
 	HttpStatus,
+	ParseArrayPipe,
 } from '@nestjs/common';
 import { DeleteResult, InsertResult, UpdateResult } from 'typeorm';
 import { AnimalService } from './animal.service';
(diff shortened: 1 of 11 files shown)


Ryno95 (changes requested):
Definite improvement to have the Batch overload! Nice to have the relations as well ,  was thinking to do it tomorrow!

Maybe a good idea to add a e2e test for the new endpoints? I thinks it's easier testing as well  as the no need to manually set-up the validators like in the unit tests!

backend/src/test_example/animal.controller.ts
```diff
@@ -31,13 +32,37 @@ export class AnimalController {
...
 
+	@Post('bulk')
+	async createBulk(
+		@Body(
+			new ParseArrayPipe({
+				items: CreateAnimalDto,
+				whitelist: true,
+				forbidNonWhitelisted: true,
+			}),
+		)
```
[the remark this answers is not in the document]
vincentvis: I tested quickly with curl, with and without extra stuff, and this is not done globally!
So `whitelist` and `forbidNonWhitelisted` are still needed here!

vincentvis:
> Definite improvement to have the Batch overload! Nice to have the relations as well , was thinking to do it tomorrow!
> 
> Maybe a good idea to add a e2e test for the new endpoints? I thinks it's easier testing as well as the no need to manually set-up the validators like in the unit tests!

added e2e tests for bulk :)

backend/src/test_example/entities/animals.entity.ts
```diff
@@ -6,5 +7,6 @@ export class AnimalEntity {
 	id: number;
 
 	@Column({ unique: true })
```
Ryno95, line 8: `@IsNotEmpy `can be removed here, then  think we can merge!

Ryno95: Already removed nice! ;D

Ryno95 (approved):

backend/src/test_example/entities/animals.entity.ts
vincentvis: done

PR #52, opened 2023-01-10
Ry/matchmaking request

Ryno95:
Basic entry points for /matchmaking-request, all using the abstract class and nothing from the db is linked, tried to keep it as simple as possible.

Files created == nest g resource matchmaking-request.

Added a unit test file and an e2e test file. With the db connection and a simple test in both cases

#41


vincentvis (approved):
Looks like the other things again... suspicious!
Lets go!

PR #84, opened 2023-01-19
Vi/cleanup 

vincentvis:
Just a big cleanup and refactor of testing imports

Removes a whole bunch of import statements at the top and just imports `AllTestingModule` which contains everything (so we only have to add stuff for testing in one spot!)
This also makes the `Test.createTestingModule` into only importing one thing!

Another "big" change:
Changed the typeorm config!
Removed the conditional part which uses the .env.test file
now just uses a different class in the typeorm config for testing than for app.module.ts...


backend/src/auth/auth.spec.ts
```diff
@@ -2,97 +2,29 @@ import { Test, TestingModule } from '@nestjs/testing';
...
-			name: 'u2',
-			password: 'p2',
-			password_confirm: 'p2',
-		},
-		{
-			name: 'u3',
-			password: 'p3',
-			password_confirm: 'p3',
-		},
-	];
```
Ryno95, line 50: The mocking was removed cause it wasn't used for anything?

backend/src/pong/game/game.service.ts
```diff
@@ -1,7 +1,7 @@
 import { Injectable } from '@nestjs/common';
 import { InjectRepository } from '@nestjs/typeorm';
 import { AbstractService } from '../../shared/abstract.service';
-import { Repository } from 'typeorm';
+import { InsertResult, Repository } from 'typeorm';
```
Ryno95, line 4: This import sneaked back in hehe

backend/src/users/user/user.controller.ts
```diff
@@ -73,6 +73,13 @@ export class UserController {
 		});
 	}
 
+	// @UseGuards(AuthGuard)
+	// // localhost:8080/api/users/id/games
+	// @Get(':id/games')
+	// async findOneWithGame(@Param('id') id: number): Promise<Game[]> {
+	// 	return this.userService.getGames(id);
+	// }
+
```
Ryno95, line 82: This sneaked back in as well, maybe with the pull from the main will resolve the last things changed.

Ryno95 (approved):
Ooooh so niiice! :D

backend/src/auth/auth.spec.ts
vincentvis: indeed

backend/src/pong/game/game.service.ts
vincentvis: pffffff
this is a "vscode is too slow" thing for me.. I open the file, it's all the same colour, I close the file....
Now I opened the file, waited 3 seconds, and the `InsertResult` became a different colour......

backend/src/users/user/user.controller.ts
vincentvis: this is probably just me sloppily merging main into this branch and picking the wrong option...

PR #85, opened 2023-01-20
Ry/user to user relationship

### LINKED ISSUE #36: USER ACCOUNT - Create User to User relationships in API

Ryno95:
Closes #36


backend/src/users/user-relationship/entities/user-relationship.entity.ts
```diff
@@ -0,0 +1,26 @@
...
+	JoinColumn,
+	OneToMany,
+	PrimaryGeneratedColumn,
+} from 'typeorm';
+import { User } from '../../user/entities/user.entity';
+
+export enum validRelationships {
+	FRIEND = 'friend',
+	BLOCKED = 'blocked',
+}
```
vincentvis, line 13: Do you know what exactly is saved in the DB here? Is it the string 'friend', or some int value?

vincentvis: also related to update-user-relationship.dto.ts btw, since there it is stored

vincentvis: asking because it might be annoying to store the string 'friend' for every user-relationship, since it's then tightly coupled with the exact word "friend"

could also be very well "out of scope" and not relevant..

vincentvis (approved):

vincentvis:
Add a user to the test beforeAll and change first test to this:

```ts
	describe('Creating relationship test', () => {
		it('should create a relationship between invite and user', async () => {
			const invite: UserRelationship = await controller.create(mockConnection);
			expect(invite.connection.length).toBe(2);
			expect(
				invite.connection.every((connectId) => connectId instanceof User),
			).toBeTruthy();
			console.log('After first:', await controller.findAll());
			const invite2: UserRelationship = await controller.create({
				connection: [2, 1],
				type: validRelationships.FRIEND,
			});
			expect(invite2.connection.length).toBe(2);
			expect(
				invite2.connection.every((connectId) => connectId instanceof User),
			).toBeTruthy();
			console.log('After second', await controller.findAll());

			const invite3: UserRelationship = await controller.create({
				connection: [3, 2],
				type: validRelationships.FRIEND,
			});
			console.log('After third', await controller.findAll());
		});
	});
```

On the last `controller.findAll()` the first UserRelationship is empty, the second one has only 1 user connected. Here's the output:
```JSON
[
  {
    "id": 1,
    "type": "friend",
    "connection": []
  },
  {
    "id": 2,
    "type": "friend",
    "connection": [
      { "id": 1, "name": "Miskruier" }
    ]
  },
  {
    "id": 3,
    "type": "friend",
    "connection": [
      { "id": 2, "name": "Tor" },
      { "id": 3, "name": "feest" }
    ]
  }
]
```

backend/src/users/user-relationship/entities/user-relationship.entity.ts
Ryno95: It is a string that i stored atm, I can be changed to another relationship with a relations table containing all valid relationship, and then just creating a relationship on id's? But seems like over-kill, could also be another ticket for crating the relationship entity, endpoints and relationships.  I feel like the enum kinda does the same thing, but indeed more tightly coupled.
Do you have other suggestions? 
And do we want to create another table? :....... (

backend/src/users/user-relationship/dto/update-user-relationship.dto.ts
```diff
@@ -0,0 +1,12 @@
+import { PartialType } from '@nestjs/mapped-types';
+import { IsEnum, IsNotEmpty } from 'class-validator';
+import { validRelationships } from '../entities/user-relationship.entity';
+import { CreateUserRelationshipDto } from './create-user-relationship.dto';
+
+export class UpdateUserRelationshipDto extends PartialType(
+	CreateUserRelationshipDto,
+) {
+	@IsEnum(validRelationships)
```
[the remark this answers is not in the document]
Ryno95: This validates the type of relationship passed is part of the defined enum, definition in user-relationship.entity,
so either friend or blocked atm.

backend/src/users/user-relationship/entities/user-relationship.entity.ts
```diff
@@ -0,0 +1,26 @@
...
+
+@Entity('user_relationships')
+export class UserRelationship {
+	@PrimaryGeneratedColumn()
+	id: number;
+
+	@OneToMany(() => User, (user: User) => user.connections) //connection
```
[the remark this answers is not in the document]
Ryno95: Um yea I think we interpret the drawing on Miro with the translation to typeorm a bit differently, or maybe exactly opposite :D 

In this case we have a user, that can have many relationships, and each relationship is between 2 users.
Let's talk about this in person tomorrow, I'll try and play around with it a bit tonight :)

I have a feeling that the ManyToOne, OneToMany might need to switch around.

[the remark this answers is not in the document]
Ryno95: Nevermind, it's just wrong, doesn't work as it should, tests are far from complete apologies.

backend/src/users/user-relationship/entities/user-relationship.entity.ts
vincentvis: I guess, as long as it's restricted by typeorm that we can only fill in 'friend' or 'blocked' (or anything for the enum) than it's fine. Then we can always display something else if we want to :P

backend/src/users/user-relationship/user-relationship.spec.ts
```diff
@@ -46,62 +48,100 @@ describe('GameInvite unit tests', () => {
...
+				source_id: user_id,
+				target_id: allUsers[0].id,
+				type: 'blocked',
+			});
+
+			const testObject: CreateUserRelationshipDto = {
+				source_id: 1,
+				target_id: 4,
+				type: 'friend',
+			};
```
vincentvis, line 134: use the variables from lines 125,126 here as well instead of hardcoded values

vincentvis (approved):

livlaming (approved):

Issue #86, opened 2023-01-22
USER ACCOUNT - Linking user_achievements table with user & achievements


Ryno95:
Create user_achievements table and link User table and achievement table

PR #88, opened 2023-01-22
Ry/achievements

### LINKED ISSUE #65: USER ACCOUNT - Achievements backend
I (Vincent) think we need to just set a fixed set of achievements. You can get a list of achievements through something like an `/achievements` endpoint and assign/delete by posting/deleting to `/achievement/{achievement_id}/{user_id}`

- [ ] Create tables
- [ ] Create endpoints
- [ ] Create relationship to user (many to many table)

Ryno95:
Nest g res achievements and constructed with by extending Abstract service. Controller fixed to work with updated service. Created test files unit and e2e with nec imports. Linking will happen in #86.

closes #65


backend/src/users/achievements/achievements.spec.ts
```diff
@@ -0,0 +1,22 @@
...
+			imports: [AllTestingModule],
+		}).compile();
+
+		service = module.get<AchievementsService>(AchievementsService);
+		controller = module.get<AchievementsController>(AchievementsController);
+	});
+
+	it('should be defined', () => {
+		expect(service).toBeDefined();
+	});
```
vincentvis, line 21: Should the controller also be defined?

vincentvis (dismissed):

backend/src/users/achievements/entities/achievement.entity.ts
```diff
@@ -0,0 +1,10 @@
+import { Column, Entity, PrimaryGeneratedColumn } from "typeorm";
+
+@Entity("achievements")
+export class Achievement {
+	@PrimaryGeneratedColumn()
+	id: number;
+
+	@Column({unique: true})
```
livlaming, line 8: is this only protecting the achievements from doubles or will it also force users to have one achievement?

vincentvis (approved):

PR #92, opened 2023-01-24
added seeding proposal

### LINKED ISSUE #80: Seed database with data on first startup of container
Many entities are dependent on having other entities existing in the database.
So we want to create something that seeds the database on first startup with some users, chats, messages, games etc.

Ryno95:
Added a simple seeding solution proposal.

Abstract service trySeed() added. 
Seeding data can be found in seed.data where defined functions return the data by using anonymous arrow functions.

A simple mechanism for only seeding once added. A hidden file .hasSeeded is created upon seeding, then its existence is checked upon every restart.

closes #80


Ryno95 (commented):
Vincent made a good point already, seeding should happen only once for each initialization of the project, so that when for example deleting a testUser or updating a testGame, the updated data should persist. So I'm trying to find a clean solution to trigger that.

backend/src/seeders/seeders.service.ts
```diff
@@ -0,0 +1,28 @@
...
+
+	shouldSeed(): boolean {
+		const shouldSeedFilePath = './dist/src/seeders/.hasSeeded';
+		if (fs.existsSync(shouldSeedFilePath)) {
+			return false;
+		} else {
+			fs.writeFileSync(shouldSeedFilePath, '');
+			return true;
+		}
+	}
```
vincentvis, line 21: Still not 100% fan of this if/else here

maybe the constructor can hold a variable that is read from 2 functions
```ts
constructor(private readonly animalService: AnimalService) {
     const shouldSeedFilePath = './dist/src/seeders/.hasSeeded';
}

shouldSeed(): boolean {
    return fs.existsSync(this.shouldSeedFilePath);
}

doneSeeding() {
    fs.writeFileSync(shouldSeedFilePath, '');
}
```

backend/src/test_example/animal.spec.ts
```diff
@@ -53,26 +48,24 @@ describe('AnimalController', () => {
...
+					name: expect.any(String),
+					id: expect.any(Number),
+				}),
+			]),
+		);
 	});
 
 	it('should get a specific animal (findOne)', async () => {
-		const animalToGet: number = 1;
+		const animalToGet = 1;
```
vincentvis, line 62: Any reason you removed the types?

backend/src/seeders/seeders.service.ts
vincentvis: Nice refactor ;-)

livlaming (approved):

backend/src/test_example/animal.spec.ts
Ryno95: no, very weird, re-added it

PR #93, opened 2023-01-26
Vi/feat intra42login

### LINKED ISSUE #4: USER ACCOUNT - 42 Api login
The user must login using the OAuth system of 42 intranet.

Information on the OAuth system of 42:
[42's API documentation](https://api.intra.42.fr/apidoc)

## todo

- [ ] Define flow of how this works (
   - which page goes where, 
   - redirects, 
   - which data do we need to send to API
   - which data do we get (and at which point)

vincentvis:
Defines 2 endpoints:
```ts
@Get('/auth/authenticate')
@Get('/auth/oauth42')
```

`/auth/authenticate` redirects to the intra login view so the user can log in (or be redirected back if they are already logged in on intra)
`/auth/oath42` is the redirect location for the previous step where "everything" happens

closes #4


backend/src/auth/dto/intra-tokendata.dto.ts
```diff
@@ -0,0 +1,26 @@
...
+	access_token: string;
+
+	@IsOptional()
+	@IsNotEmpty()
+	refresh_token?: string;
+
+	@IsOptional()
+	@IsNotEmpty()
+	scope?: string;
+
```
livlaming, line 18: curious what the scope does or mean in this one?

backend/tsconfig.json
```diff
@@ -9,7 +9,7 @@
 		"target": "es2017",
 		"sourceMap": true,
 		"outDir": "./dist",
-		"baseUrl": "../../",
+		"baseUrl": "./",
```
livlaming, line 12: what impact would this one have? (just curious)

livlaming (approved):
not sure I fullt grasp it but as far as I can understan it looks good

backend/tsconfig.json
vincentvis: something with vscode (or other editor maybe?) generating correct `import` paths, but so far non of the settings seem to work reliably... so I change this around sometimes and forget to not commit/push this change.. so it ends up in the repo xD

backend/src/auth/dto/intra-tokendata.dto.ts
vincentvis: from:  https://api.intra.42.fr/apidoc/guides/web_application_flow
> A space separated list of [scopes](https://api.intra.42.fr/apidoc/guides/web_application_flow#scopes). If not provided, scope defaults to an empty list of scopes for users that don't have a valid token for the app. For users who do already have a valid token for the app, the user won't be shown the OAuth authorization page with the list of scopes. Instead, this step of the flow will automatically complete with the same scopes that were used last time the user completed the flow.

We don't use it, but it exists in the response of the api..

backend/src/auth/auth.service.ts
```diff
@@ -1,10 +1,124 @@
...
+		return await fetch('https://api.intra.42.fr/oauth/token', {
+			method: 'POST',
+			mode: 'cors',
+			body: data,
+		})
+			.then((response) => response.json())
+			.then(async (data) => data);
+	}
+
+	async newUserFromIntra(userData: any) {}
```
Ryno95, line 69: Not sure if this bad boy was still intended for something?

backend/src/auth/auth.service.ts
```diff
@@ -1,10 +1,124 @@
...
+		return await fetch('https://api.intra.42.fr/v2/me', {
+			headers: {
+				Authorization: `Bearer ${data.access_token}`,
+			},
+		})
+			.then((response) => response.json())
+			.then((userData) => userData);
+	}
+
+	async exchangeCodeForToken(code: string, state: string) {
```
Ryno95, line 51: state is not being used

Ryno95 (approved):
Looks good!

Tried doing weird things to try and get something to fail, seems solid!

Amazing naming ;)

Two general questions:

- Do we also want to get the pfp from intra as default avatar?
- crypto === bcrypt? SSBD?  <=== Same same but different

backend/src/auth/auth.service.ts
vincentvis: yes..

PR #103, opened 2023-01-27
Hashed password chat

livlaming:
sorry I forgot the Li/ before the branch name!
small pr :)!

### DIFF
--- a/backend/src/chats/chat/chat.controller.ts
+++ b/backend/src/chats/chat/chat.controller.ts
@@ -12,14 +12,18 @@ import { ChatService } from './chat.service';
 import { CreateChatDto } from './dto/create-chat.dto';
 import { UpdateChatDto } from './dto/update-chat.dto';
 import { Chat } from './entities/chat.entity';
+import * as bcrypt from 'bcrypt';
 
 @Controller('chats')
 export class ChatController {
 	constructor(private readonly chatService: ChatService) {}
 
 	@Post()
-	create(@Body() createChatDto: CreateChatDto) {
-		// use bcrypt to hash passwords!
+	async create(@Body() createChatDto: CreateChatDto) {
+		if (createChatDto.hasOwnProperty('password')) {
+			const hashed = await bcrypt.hash(createChatDto.password, 11);
+			createChatDto.password = hashed;
+		}
 		return this.chatService.insert(createChatDto);
 	}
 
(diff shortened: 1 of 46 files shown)


backend/src/chats/chat/chat.controller.ts
```diff
@@ -12,14 +12,18 @@ import { ChatService } from './chat.service';
...
 	constructor(private readonly chatService: ChatService) {}
 
 	@Post()
-	create(@Body() createChatDto: CreateChatDto) {
-		// use bcrypt to hash passwords!
+	async create(@Body() createChatDto: CreateChatDto) {
+		if (createChatDto.hasOwnProperty('password')) {
+			const hashed = await bcrypt.hash(createChatDto.password, 11);
+			createChatDto.password = hashed;
+		}
```
Ryno95, line 26: Same as  #102 not possible to create a chat without a password, in DTO = optional in entity mandatory. 
Do we want to be able to make chats without a pw?  If so we can add it as an optional in the entity.

Ryno95 (commented):
Looks good, only comment is the same as in #102 :)

Ryno95 (commented):
Also, there is no validation for the visibility at this point, if we are doing private/public, you can use an enum  or something like this: 
``` typescript
const validVisibility = ['public', 'private'];

export class Chat {

    @IsIn(validVisibility)
  visibility: string
}
```

backend/src/chats/chat/chat.controller.ts
vincentvis: this is a better spot for this xD But I had already responded in the other PR as well.
but copy pasta here for completeness sake

Changing the entity password to @Column({ nullable: true }) solves this

backend/test/chat.e2e-spec.ts
```diff
@@ -5,10 +5,21 @@ import { ChatController } from '../src/chats/chat/chat.controller';
...
 describe('chat e2e', () => {
 	let app: INestApplication;
 	let chatController: ChatController;
+	let messageService: MessageService;
+	let messageController: MessageController;
+
+	const testMessages: CreateMessageDto[] = [
+		{ sender_id: 1, chat_id: 1, content: 'hello' },
+		{ sender_id: 2, chat_id: 1, content: 'world!' },
+	];
```
vincentvis, line 21: This is added but not used in the e2e test

(same for MockChat a few lines below this)

vincentvis (changes requested):
Needs an update to allow chats without password, but other than that good :)

backend/src/chats/chat/chat.controller.ts
livlaming: alright!

vincentvis (approved):

Ryno95 (approved):
Nois! MEEEERGE THAT BRANCH!!! :D

PR #104, opened 2023-01-30
FRONTEND - Vi/frontend navigation

### LINKED ISSUE #95: Create navigation html/css/vue + empty tabs for content

vincentvis:
First frontend PRRRRRRRR

some cleanup
some default css stuff

navigation bar #95 
and empty tabs! (well.. with a title..)

closes #95


compose.yml
```diff
@@ -1,18 +1,19 @@
 services:
     frontend:
-        container_name: "frontend_${COMPOSE_PROJECT_NAME:-ts}"
+        container_name: 'frontend_${COMPOSE_PROJECT_NAME:-ts}'
```
Ryno95, line 3: Appending the -ts, does that solve compiling across multiple versions of the project? I think you explained it before but not specifically what changed

frontend/src/components/__tests__/not-a-test.spec.ts
```diff
@@ -0,0 +1,7 @@
+import { describe, it, expect } from 'vitest';
+
+describe('Not a Test', () => {
+	it('is not testing anything', () => {
+		expect(true).toBeTruthy();
```
Ryno95, line 5: 🤣

Ryno95 (approved):
Curiousity question... Why do we have 7 of the pong images? I could only find two of them being used in `frontend/public/site.webmanifest`

Thanks for setting up the template for the frontend, feels like 80% of the project is now done 🤓😬🤫

compose.yml
vincentvis: this adds a `-ts` suffix when the COMPOSE_PROJECT_NAME is not set (which it probably isnt.)
so if you run compose like this:
`COMPOSE_PROJECT_NAME=CHECKPR docker compose up --build`
it appends `CHECKPR` to all the containers to stop it from complaining about "container already in use"

vincentvis:
> Curiousity question... Why do we have 7 of the pong images? I could only find two of them being used in `frontend/public/site.webmanifest`
> 
> Thanks for setting up the template for the frontend, feels like 80% of the project is now done 🤓😬🤫

Most of those images are automagically selected by the browser/device that wants them
This is all generated by [realfavicongenerator](realfavicongenerator.net/) and I just copy paste the files etc from there :)

frontend/src/assets/main.css
```diff
@@ -1,35 +1,59 @@
 @import './base.css';
```
livlaming, line 2: not sure where to place this comment, the menu looks amazing!
If we will be using the lined square maybe its nice to make is a bit more in the same ratio as the design? (a bit more squarisch but not fully)

vincentvis: which lined square? I did nothing with a lines square 🙃

frontend/public/browserconfig.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
```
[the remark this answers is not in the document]
vincentvis: it's just an icon for when you add it to your homescreen on an ios device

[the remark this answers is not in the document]
livlaming: haha alright fun!

frontend/src/assets/main.css
livlaming: Screen Shot 2023-01-30 at 3 26 47 PM

livlaming: deze rechthoek rondom this is a login page

vincentvis: I guess that is indeed part of this first setup thingy :P 
I honestly just made a box, not really worrying about the size, but I adjusted it now to be 4/5 aspect ratio (which is close to what it is in penpot)

livlaming: perfect!

PR #105, opened 2023-01-30
Ry/profiles list

### LINKED ISSUE #98: Create Profile List component

Ryno95:
Changes implemented:

- We can now get avatars from public in the backend by using the endpoint/src: `/api/avatars/${avatarImgName}`
- We now have a List of users in profiles, that we can, block/unblock, friend/unfriend, it routes to chat and singleProfileView with the id of the user clicked, passed as a param.
- If a user is blocked, other options are now greyed out, if other implementation is needed we can open a ticket for that.

Possible things that still need to be thought about:

- Saving the current relationship somehow, I played around with it a bit, but it is deceptively complicated (ticket).
- Find a better way to collect blocked user, or friends or to check if someone isBlocked or a friend.

closes #98


backend/public/index.html
```diff
@@ -0,0 +1 @@
+HAALO!
```
vincentvis, line 1: Do we need this file? (maybe it's required.. who knows 🤷) but maybe then different file contents?

backend/src/auth/auth.guard.ts
```diff
@@ -9,6 +9,7 @@ export class AuthGuard implements CanActivate {
 		context: ExecutionContext,
 	): boolean | Promise<boolean> | Observable<boolean> {
 		// return true;
+
```
vincentvis, line 12: this file will have an amazing edit history in the end 😆

frontend/src/components/profileList/Avatar.vue
```diff
@@ -0,0 +1,25 @@
+<template>
+	<div class="c_media">
+		<a
+			href="#"
```
vincentvis, line 4: I realized now that this should probably become a router-link element. I'll have a look later how to restructure the html.. (also for other A elements that actually go somewhere..)

frontend/src/components/profileList/BlockUser.vue
```diff
@@ -0,0 +1,40 @@
+<template>
+    <div v-if="isBlocked(userId) === true">
+       <a href="#" v-on:click.prevent="updateRelationship(userId, ValidRelationships.NONE)" >Unblock</a>
```
vincentvis, line 3: did you see [$emit()](https://vuejs.org/guide/components/events.html#emitting-and-listening-to-events)?
Not sure if that applies here, as always, I'm in the train and can't build a new container...

frontend/src/components/profileList/ListRow.vue
```diff
@@ -0,0 +1,90 @@
...
+			await router.push({ name: 'chat', params: {id: userId} })
+		},
+		async routeToProfile(userId: number) {
+			await router.push({name: 'profile', params: {id : userId}})
+		},
+	},
+});
+</script>
+
+<style>
```
vincentvis, line 83: should this have a "scoped" attribute? Or does that break it since the class is not used in this component but in a sub-component?

frontend/src/components/profileList/ListRow.vue
```diff
@@ -0,0 +1,90 @@
...
+	props: {
+		user: {
+			type: Object as PropType<User>,
+			required: true,
+		},
+		online: Boolean,
+	},
+	methods: {
+		async routeToChat(userId: number) {
+			console.log(`Starting chat with ${userId}`);
```
vincentvis, line 73: Will there be logic here to check if chat exists, and if not, make the chat? Or should that be done at some other point?

And should we create a ticket for that? 🎟️

frontend/src/router/index.ts
```diff
@@ -57,7 +57,7 @@ const router = createRouter({
 			component: () => import('../views/NewChannelView.vue'),
 		},
 		{
-			path: '/profile',
+			path: '/profile/:id',
```
vincentvis, line 60: Would we need both `/profile/` and `/profile/:id`
one for your own profile and one for others their profile?

Or will `/settings` (/me or whatever it will be) be "editing your own profile" and `/profile/:id` with your own ID be the page where you can view your own profile?

frontend/src/stores/userStore.ts
```diff
@@ -59,7 +57,7 @@ export const useUserStore = defineStore('users', {
 		async refreshAllUsers() {
 			try {
 				const data = await getRequest('users');
-				this.allUsers = data.data;
+				this.allUsers = data.data.filter((user: User) => (user.id !== this.me.id));
```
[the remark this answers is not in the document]
vincentvis: I understand now that this is to view "all users" on the /profiles page

Point to discuss:
also show the logged in user their profile
this way a user can see their own profile in the same way it would also be visible to others

frontend/src/components/profileList/ListRow.vue
vincentvis: On second thought, and better look, this does not seem to be used (the .avatar)
so can be removed

frontend/src/components/profileList/Avatar.vue
vincentvis: This works: _(you have to pass the id to the Avatar component)_

```vue
<router-link to="`/profiles/${id}`">
	<div
		class="c_media__asset"
		:class="{
			'c_asset--online': isOnline,
			'c_asset--offline': !isOnline,
		}"
	>
		<div class="c_asset__circle">
			<img :src="`/api/avatars/${avatar}`" alt="" />
		</div>
	</div>
</router-link>
```

backend/public/index.html
Ryno95: added option `index: false`, otherwise it looks for a default index.html page to returns, otherwise it throws. removed index.html.

frontend/src/components/profileList/Avatar.vue
Ryno95: Discussed in person, routing via script instead of html element

frontend/src/components/profileList/ListRow.vue
Ryno95: Added tcket

frontend/src/components/profileList/BlockUser.vue
Ryno95: Yes, I din't understand it completely, if it's an improvement on the current implementation, please let me know so I can dive into it, like I mentioned, there are 15 ways to do things, and if there's one that works and that i understand I run with that until further notice :)

frontend/src/components/profileList/ListRow.vue
Ryno95: Indeed un-used, now removed

frontend/src/router/index.ts
Ryno95: We can maybe add an option to "setting for viewing your own profile? Don't mind if it's with `/:id` or without, I think we can just reuse the `/:id` path? Not sure that it makes sense to a have completely diff.

frontend/src/stores/userStore.ts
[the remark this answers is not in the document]
Ryno95: Uuummmmm apples :D

See above comment about `view my profile`.

frontend/src/components/profileList/BlockUser.vue
vincentvis: I believe it's meant for emitting events to the parent component/view
so if the ListRow component has an "updateRelationship" method, you could emit 
```vue
<a href="#" @click.prevent="$emit('updateRelationship', userId, id_of_new_relation)">
```
and have the parent component handle the updating

But I also realize now that the updateRelationship is a method/action from the userStore, so not sure how that ties together..


just leave it as is I guess :P

backend/public/index.html
livlaming: haah love this

Ryno95:
Okey incoming updates:

- Removed 666 api calls for relationship. now getting the relationshipType once or each user.
- Tried removing the index.html with specific settings which should allow that, buuuuuuut no luck on that yet, so re-added it with no body.
- Removed unused styles in listRow

PR #109, opened 2023-02-02
Ry/seeding this farm

Ryno95:
Seeded the following:

- Users
- Games
- user Relationships
- User chats (will prob change with new relationships)

Not finalized, not perse the best way, but it's something.


backend/src/seeders/seeders.service.ts
```diff
@@ -19,8 +35,85 @@ export class SeederService {
...
+				type: validRelationships.FRIEND,
+			},
+			{
+				source_id: allUsers[0].id,
+				target_id: allUsers[2].id,
+				type: validRelationships.BLOCKED,
+			},
+		];
+		await this.userRelService.save(rel);
+	}
```
vincentvis, line 108: Depending on which branch will be merged first, this might fail (if new relationships are okay and pushed)
But I can take the seeding of that part after the branches are merged, so maybe just comment this out for now?

vincentvis (approved):

PR #110, opened 2023-02-03
Vi/frontend elements

### LINKED ISSUE #106: Create html+css elements for frontend to be reused in components

vincentvis:
Reusable elements for the frontend
still have to be put into components, and split from the one big css file, but should give good starting point to build with.

might still needs some fixes for different screen sizes, especially for chat/messages related views

look at pages:
/chat
and /elements-to-reuse (added this just so we have a place to put it, link will be removed from navigation later)

closes #106


livlaming (approved):
Looks great!!

just two thoughts:

- one grey line suggestion
- size of text area suggestion: maybe its nice to make the type area of the chatroom a little bit higher see example:
Screen Shot 2023-02-06 at 1 02 42 PM

elements zien er ook top uit, alleen misschien die grijze lijn, ie png
Screen Shot 2023-02-06 at 1 10 28 PM

Ryno95 (approved):
Looks good! 

One question, should the text box in the chat be resizable?

PR #127, opened 2023-02-07
Ry/login fe

### LINKED ISSUE #63: USER ACCOUNT - login view
A view that shows a page to login
Most prominent choice should be "login with intra 42 API"
Below that a smaller link with "login with email" which should show a small form (username/password + submit)

### LINKED ISSUE #112: Login Component / View

Ryno95:
Added logic LoginView.vue using the reusable components ;)

Um Tried to stick to the wireframe design.  Only thing thats off is the font-weight, gimme a shout if that's something important.

Logging in with intra redirects succesfully to api and then back to the HomeView of the frontend. 

Logging in with username and pw redirects to homeview as well.

I added a types/Constants.ts where i export the backend api url.

Clicking in register should also redirect to the incoming RegisterView.

closes #63


frontend/src/views/LoginView.vue
```diff
@@ -1,13 +1,51 @@
 <template>
 	<div class="page_box_wrapper">
-		<div class="page_box">
-			<h1>This is a login page</h1>
-		</div>
+        <h1> HI WANNA PLAY PONG? </h1>
```
vincentvis, line 3: you're missing the `<div class="page_box">` around the content

```html
     <div class="page_box">
          <!-- Content should be here -->
     </div>
</div>
```

frontend/src/views/LoginView.vue
```diff
@@ -1,13 +1,51 @@
...
+                    <p>
+                        <a href="#" v-on:click.prevent="userStore.login(loginForm)">Login</a>
+                        /
+                        <a href="/register">Register</a>
+                    </p>
+                    <p>
+                        <a :href="redirectUrl">Sign in with 42</a>
+                    </p>
+                </div>
+            </div>
```
vincentvis, line 19: Sooo, here I was wrong
Login with the <a> element is annoying (no action on enter)

this is my suggestion:

```vue
<form
	method="post"
	action=""
	class="c_block c_form_group"
	@submit.prevent="userStore.login(loginForm)"
>
<!-- inputfields -->

<!-- instead of this:
<a href="#" v-on:click.prevent="userStore.login(loginForm)">Login</a> -->
<input class="link_button" type="submit" value="Login" />
</form>
```

and this added to main.css somewhere
```css
.link_button {
	-webkit-appearance: none;
	appearance: none;
	background: none;
	border: 0px;
	padding: 0;
}
```

frontend/src/stores/userStore.ts
```diff
@@ -17,12 +18,10 @@ export const useUserStore = defineStore("users", {
...
-    async login() {
-      const loginData = {
-        name: "renoster",
-          password: 'R'
-        }
-        await postRequest("login", loginData);
+    async login(loginForm: LoginForm) {
+      await postRequest("login", loginForm);
+      await this.refreshMe();
+      await router.push("/")
```
vincentvis, line 24: If the user inputs wrong info, or nothing, there is an error in the console, but nothing else happens.
Do you want catching this error to be a separate ticket and handled later, or do you want to implement that now?

vincentvis: it would also change something in the html btw, since you'd want to show an error message visually. So maybe a separate task is nice.. but you pick!

vincentvis: Could also maybe add an options, that if a username is entered that exists as an intra_user we show a message "did you mean to login through the API?" or something like that..

vincentvis (changes requested):

frontend/src/stores/userStore.ts
Ryno95: added a simple alert in the try catch block of userStore->login() and added a ticket #130

frontend/src/views/LoginView.vue
```diff
@@ -1,22 +1,33 @@
...
+					class="c_block c_form_group"
+					@submit.prevent="userStore.login(loginForm)"
+					>
+					
+					<InputField label="Username" placeholder="username" v-model="loginForm.name"/>
+					<InputField label="Password" placeholder="password" v-model="loginForm.password"/>
+					
+					<div class="c_block c_split">
+						<p>
+							<input v-on:submit="userStore.login(loginForm)" class="link_button" type="submit" value="Login" />
```
vincentvis, line 20: is this second `v-on:submit=""` needed here? I thought it works without this (since the button submits the form, and the form already has a submit action)

vincentvis: I checked, the v-on on the input is not needed !

frontend/src/views/LoginView.vue
```diff
@@ -1,13 +1,66 @@
...
+							/
+							<a href="/register">Register</a>
+						</p>
+						<p>
+							<a :href="redirectUrl">Sign in with 42</a>
+						</p>
+					</div>
```
livlaming, line 27: looks nice, would be cool to handle errors when the username and password are wrong or empty for example

frontend/src/assets/main.css
```diff
@@ -478,3 +478,16 @@ h1 {
...
 	}
 }
+
+.link_button {
+	-webkit-appearance: none;
+	appearance: none;
+	background: none;
+	border: 0px;
+	padding: 0;
+	font-style: italic;
```
vincentvis, line 488: could you add this as well?
```css
cursor: pointer;
```

vincentvis (approved):

livlaming (commented):
cool only one comment about the errorhandling, and something with the square box but I sw you just fized that :), great example :~)

frontend/src/assets/main.css
Ryno95: I had it, then removed, as it seemed not to make a difference, maybe just very slow hot reloading on my ancient machine :D

frontend/src/views/LoginView.vue
Ryno95: Yes indeed! At the moment it does throw an alert for non matching usern or passw, or empty. I opened a new ticket for a more verbose error handling

livlaming: wonderfull!

PR #128, opened 2023-02-07
Vi/endpoint relations and more

### LINKED ISSUE #122: Create endpoints for getting entities with and without their relationships
Maybe you can specify in the query params which relationships you want?

`../api/users?friends=true,matchmaking_requests=true`

OR
better?

add optional DTO to get/get_one with relationships

vincentvis:
Main things done:

- added DTO's to tell user/chat endpoint which relationships you want to load
It can either be done through sending a body, or through query parameters

- added timestamps to messages for seeding (since they were all exactly send at the same microsecond, which is bit weird..)

closes #122


backend/src/chats/chat-user-permissions/entities/chat-user-permission.entity.ts
```diff
@@ -43,4 +45,10 @@ export class ChatUserPermission {
 	})
 	@JoinColumn({ name: 'permission_id' })
 	permissions: Permission[];
+
+	@CreateDateColumn()
+	created_at: Date;
+
+	@UpdateDateColumn()
+	updated_at: Date;
 }
```
livlaming, line 54: not sure why you need time and date for the permissions? but you prob have a good reason :)

backend/src/chats/chat/chat.controller.ts
```diff
@@ -7,18 +7,29 @@ import {
...
 @Controller('chats')
 export class ChatController {
 	constructor(private readonly chatService: ChatService) {}
 
+	private readonly defaultRelationships = { has_users: true };
+	private readonly defaultOrder: FindOptionsOrder<Chat> = {
+		name: 'ASC',
+		messages: {
+			created_at: 'ASC',
+		},
```
livlaming, line 30: created at ASC? me is confused (:

backend/src/chats/chat/chat.controller.ts
```diff
@@ -29,15 +40,38 @@ export class ChatController {
...
+	findAll(
+		@Query() userRelationsQuery?: ChatRelationsQueryDto,
+		@Body() userRelationsBody?: ChatRelationsBodyDto,
+	) {
+		const userRelationsDto = Object.keys(userRelationsBody ?? {}).length
+			? userRelationsBody
+			: Object.keys(userRelationsQuery ?? {}).length
+			? userRelationsQuery
+			: this.defaultRelationships;
+
```
Ryno95, line 52: Vincent, I understand what the result of this block of code is, but I have no idea how it is achieved.... could solve some of this abstraction for me with a short explanation please?

livlaming (commented):
whaaa hard work! a lot of work! well done

backend/src/chats/chat/chat.controller.ts
Ryno95: wait i get it, as bunch of ternary expression, if URbody, map and use that, else if URquery map and use that, else default, right?

backend/src/users/user/user.controller.ts
```diff
@@ -21,16 +22,31 @@ import * as bcrypt from 'bcrypt';
...
 export class UserController {
 	constructor(private userService: UserService) {}
 
+	private readonly defaultRelationships = {};
+	private readonly defaultOrder: FindOptionsOrder<User> = {
+		name: 'ASC',
+		in_chats: {
+			created_at: 'ASC',
+		},
+	};
```
Ryno95, line 48: Not used atm

Ryno95 (approved):
Super cool addition!

backend/src/chats/chat/chat.controller.ts
vincentvis: this is for ordering when doing a query
so the default order for the messages is on the `created_at` column in ASCending order

backend/src/chats/chat/chat.controller.ts
vincentvis: yes

i know, i hate it too.. but i gave up in live at some point...
I wanted to make a helper function that just gets the 3 options as params and figures it out with if/else. 
But I feel like those kind of helpers should live in their own file in TS/nestjs, or at least not be a method of the controller. So I gave up on the thought of figuring out where to put it and what to name it...

backend/src/chats/chat-user-permissions/entities/chat-user-permission.entity.ts
vincentvis: ehhhhhhhhh yes.. there was something.. I kinda forgot..
but it doesn't hurt to have it, and it's auto generated/auto updated etc, so it should not be in the way of anything even if I never remember what it was for..

livlaming: haha alright

backend/src/chats/chat/chat.controller.ts
livlaming: cool!

PR #129, opened 2023-02-08
Li/user account

livlaming:
register pageeeee

### DIFF
--- a/frontend/src/views/RegisterView.vue
+++ b/frontend/src/views/RegisterView.vue
@@ -8,12 +8,22 @@
 				class="c_block c_form_group"
 				@submit.prevent="userStore.register(registerForm)"
 			>
-				<InputField label="username" v-model="registerForm.name"/>
-				<InputField label="password" v-model="registerForm.password"/>
-				<InputField label="confirm password" v-model="registerForm.password_confirm"/>
+				<InputField label="username" v-model="registerForm.name" />
+				<InputField label="password" v-model="registerForm.password" />
+				<InputField
+					label="confirm password"
+					v-model="registerForm.password_confirm"
+				/>
 				<div class="c_block c_split">
-					<p><a href="/login">back</a> </p>
-					<p><input class="link_button" type="submit" value="register" /></p>
+					<p><a href="/login">back</a></p>
+					<p>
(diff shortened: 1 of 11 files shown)


frontend/src/views/RegisterView.vue
```diff
@@ -0,0 +1,65 @@
...
+				<InputField
+					label="confirm password"
+					v-model="registerForm.password_confirm"
+				/>
+				<div class="c_block c_split">
+					<p><a href="/login">back</a></p>
+					<p>
+						<input
+							class="link_button"
+							action="/settings"
```
vincentvis, line 22: the `input` element does not have an `action` attribute in html (does this do something?)

frontend/src/views/RegisterView.vue
```diff
@@ -0,0 +1,65 @@
...
+							type="submit"
+							value="register"
+						/>
+					</p>
+				</div>
+				<div v-if="userStore.errors.length">
+					<p v-for="error in userStore.errors">
+					<p style="color:red;">!! {{ error }}</p>
```
vincentvis, line 30: sooo in html/css inline styles are considered bad :P 
it's generally better to avoid using `style=""` on any element unless otherwise impossible
better to use something like `class="error"` and add to the `<style>` section `.error { color: red; }`

you open 2 paragraph elements inside of each other
so that is repeated for every error, 3 errors become this:
```html
<p>
    <p class="error">error 1</p>
</p>
<p>
    <p class="error">error 2</p>
</p>
```
You can move the class (/style) attribute together with the v-for:
```vue
<p v-for"error in userStore.errors" class="error">!! {{ error }}</p>
```

vincentvis (changes requested):

Ryno95 (approved):
Looks good, only one discussion point, do we need to have like minimum lengths and or complexity for passwords and or usernames?

frontend/src/views/RegisterView.vue
livlaming: haha yes you are very right! I'll remove this, this came from absolute desperation  for an answer

vincentvis (approved):

PR #139, opened 2023-02-13
Li/profile component

### LINKED ISSUE #99: Create Profile component (viewing others profile)

livlaming:
hola,
the wins and losses need to bee connected when we have the endpoints, but for now Id love to know if you agree with the set up.

achievements should look neater.

Ps test with:

http://localhost:8080/profile/userIdThatActuallyExists!

example:

http://localhost:8080/profile/15

closes #99


backend/src/users/user-achievements/user-achievements.controller.ts
```diff
@@ -32,11 +32,16 @@ export class UserAchievementsController {
...
+	// @Get()
+	// async findAll() {
+	// 	return this.userAchievementsService.findAll({
+	// 		relations: { users: true, achievements: true },
+	// 	});
+	// }
+
+	@Get(':id')
+	async findAll(@Param('id') id: number) {
+		return this.userAchievementsService.findAll({ where: { user_id: id } });
```
vincentvis, line 44: this now does not really do "all user-achievements" anymore, right?
it does "all user-achievements of user with id = id"

so maybe leaving it here it was
and doing this
- user controller `@Get(':id/achievements')` (to do what you do here, get achievements of a single user)
- me controller `@Get('achievements') (to get achievements of current user)

frontend/src/stores/userStore.ts
```diff
@@ -12,27 +12,37 @@ export const useUserStore = defineStore('users', {
...
+
+		// async getUser(id: number){
+		// 	try{
+		// 		await getRequest('/users/${id}');
+		// 		return User[];
+		// 	}
+		// 	catch (e) {
+		// 		alert('Invalid user');
+		// 	}
+		// },
```
vincentvis, line 40: ?

frontend/src/views/ProfileView.vue
```diff
@@ -1,10 +1,74 @@
...
+			user: {} as User,
+			userAchievement: [] as UserAchievement[],
+			achievement: [] as Achievement[],
+		};
+	},
+	async created() {
+		this.user = (await getRequest(`users/${this.profile_id}`)).data;
+		if (!this.profile_id) {
+			dataArray.value = [
+				{ left: 'this profile', right: 'does not exist' },
```
vincentvis, line 49: this is not going through the userStore, is there a specific reason? Or is that a "later" thing?

frontend/src/views/ProfileView.vue
```diff
@@ -1,10 +1,74 @@
...
+			dataArray.value = [
+				{ left: 'intra name', right: this.user.name },
+				{ left: 'member since', right: this.user.created_at },
+				{ left: 'wins', right: 1 }, //needs to be connected
+				{ left: 'losses', right: 190 }, //needs to be connected
+				{
+					left: 'achievements',
+					right: this.user.achievements.map((ach) => {
+						return ach.name;
+					}),
```
vincentvis, line 61: This makes the achievements into an array.. and displays it like this:

Screen Shot 2023-02-13 at 3 21 53 PM

So maybe for now add a join with a comma after the map? (`map().join(', ')`)

backend/src/users/user/dto/create-user.dto.ts
```diff
@@ -1,4 +1,5 @@
 import { IsBoolean, IsNotEmpty, IsOptional } from 'class-validator';
+import { Transform } from 'class-transformer';
```
Ryno95, line 2: Unused import

frontend/src/stores/userStore.ts
Ryno95, line 40: Should this still be removed?
Also, if this is going to be used, this should return a single user as it's pinging the findOne controller :)

frontend/src/types/User.ts
```diff
@@ -37,3 +37,8 @@ export interface User {
 	created_at: Date;
 	updated_at: Date;
 }
+
+export interface UserAchievement {
+	user_id: number;
+	achievement_id: number;
+}
```
Ryno95, line 44: I don't this is being used, or will be used, can be deleted

frontend/src/views/ProfileView.vue
```diff
@@ -1,10 +1,79 @@
...
 
+<style scoped></style>
+
 <script lang="ts">
-</script>
\ No newline at end of file
+import { defineComponent } from 'vue';
+import VerticalAvatarAndUserName from '@/components/user-info/VerticalAvatarAndUserName.vue';
+import OverviewWithMidline from '@/components/overviews/OverviewWithMidline.vue';
+import { ref, onMounted, type Ref, defineProps } from 'vue';
```
Ryno95, line 19: Unused imports from our figuring out stage: onMounted and defineProps

frontend/src/views/ProfileView.vue
```diff
@@ -1,10 +1,79 @@
...
+import type { OverviewArray } from '@/types/OverviewArray';
+import { getRequest } from '@/utils/apiRequests';
+import type { User } from '../types/User';
+
+let dataArray: Ref<OverviewArray[]> = ref([]);
+let user: User = {} as User;
+
```
Ryno95, line 26: these can be removed, ref is not needed as we are not mutating any data. Also the user declared here is not being used.
You can put the dataArray in the data func like this;
``` typescript
data() {
    dataArray: [] as OverviewArray[],
}
```

frontend/src/views/ProfileView.vue
```diff
@@ -1,10 +1,79 @@
...
+		return {
+			user: {} as User,
+		};
+	},
+	async created() {
+		this.user = (await getRequest(`users/${this.profile_id}`)).data;
+		// await useUserStore().refreshAllUsers();
+		// this.user = useUserStore().allUsers.filter((user: User) => (Number(user.id) === Number(this.profile_id)))[0];
+		if (!this.profile_id) {
+			dataArray.value = [
```
Ryno95, line 46: if you change the dataArray to be intanciated in the `data()` then dataArray.value becomes this.dataArray.

frontend/src/views/ProfileView.vue
```diff
@@ -1,10 +1,79 @@
...
+		}
+	},
+
+	setup() {
+		return {
+			dataArray,
+			name: user.name,
+		};
+	},
+});
```
Ryno95, line 73: This can be removed

Ryno95 (approved):
A few refactoring comments :)

frontend/src/views/ProfileView.vue
```diff
@@ -1,10 +1,79 @@
...
+	props: {
+		profile_id: String,
+	},
+	data() {
+		return {
+			user: {} as User,
+		};
+	},
+	async created() {
+		this.user = (await getRequest(`users/${this.profile_id}`)).data;
```
Ryno95, line 42: I would move this line into the else block, now you are doing a get request using the id, and then checking if the id exists

frontend/src/views/ProfileView.vue
```diff
@@ -35,19 +31,18 @@ export default defineComponent({
...
 	async created() {
-		this.user = (await getRequest(`users/${this.profile_id}`)).data;
-		// await useUserStore().refreshAllUsers();
-		// this.user = useUserStore().allUsers.filter((user: User) => (Number(user.id) === Number(this.profile_id)))[0];
-		if (!this.profile_id) {
-			dataArray.value = [
-				{ left: 'this profile', right: 'does not exist' },
-			];
+		await useUserStore().refreshAllUsers();
+		const filteredUsers = useUserStore().allUsers.filter((user: User) => Number(user.id) === Number(this.profile_id));
```
vincentvis, line 40: you can use `find` here instead of `filter`
find returns 1 instance, not an array, and stops at the first "found" instance. If it is not found the return value is 'undefined'
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find

vincentvis (approved):

vincentvis (approved):

PR #143, opened 2023-02-14
Ry/game elements

### LINKED ISSUE #69: Create base of game (Canvas, Ball, Bars)
Calculations of game elements their height/width/radius should be relative to play area (percentage based, but converted back to pixels for canvas drawing)

Ryno95:
Added basic html elements to game.

Drawing bars, ball and mid line using canvas methods.

closes #69


frontend/src/views/GameView.vue
```diff
@@ -1,51 +1,156 @@
...
+					lineWidth,
+					lineHeight
+				);
+			},
+			
+		//  https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/arc
+		drawBall() {
+			const startingX: number = this.width / 2 - 100;
+			const startingY: number = this.height / 2 - 100;
+			const radius: number = 15;
```
vincentvis, line 84: If we want these fully relative to canvas, the `- 100` part should also some percentage based value (`this.width/100*10` for 10% or something like that)

frontend/src/views/GameView.vue
```diff
@@ -1,51 +1,156 @@
...
+			)
+				// this fills the above defined arc/circle
+				this.context.fill()
+			},
+				
+		// https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/fillRect
+		drawPaddles() {
+			let widthStart: number = 0;
+			const lineWidth: number = 15;
+			const lineHeight: number = this.height / 5;
```
vincentvis, line 103: the lineHeight now becomes 20% of the canvas height.
Which is, to me, not the clearest way to calculate this (as in, this does not immediately shout "i am an percentage" to me (`this.height / 100 * 20` does that more, but is annoying)
so maybe we can add some computed property that is 1% of the width and another one that is 1% of the height? (just a suggestion) because then you can just always use that and multiply by the percentage wanted.


also:
I guess the paddle-height multiplier / divisor should be a variable as well (if we ever want to make the paddles change size when you make a point or something like that..) Not sure if we should already consider different paddle heights per player.......................

vincentvis (approved):

frontend/src/views/GameView.vue
Ryno95: Indeed, adde methods width/heightPercentage(perc: number) that returns the wanted percentage value

frontend/src/views/GameView.vue
Ryno95: changed to percentage values.

Added types for later use. All moving elements will be variables, that only starts making sense when movement starts as now it will be a declaration and then filling in values, cluttering unnecessarily . For now this seems clearer as it's just const values being drawn once.

vincentvis (approved):

frontend/src/views/GameView.vue
vincentvis: perfect!

frontend/src/views/GameView.vue
vincentvis: good points :) const values for now are fine!

frontend/src/views/GameView.vue
```diff
@@ -1,51 +1,167 @@
...
+}
+
+.gameHeader{
+	font-weight: bold;
+	display: flex;
+	/* font size changes on window HEIGHT! */
+	font-size: 2.5vh; 
+	padding-left: 5%;
+	padding-right: 5%;
+	padding-top: 4%;
```
livlaming, line 160: padding top: 2.4% ? 

this header with scores and 'PONG' seem a little low, 
I think it would be nice if its at least at the same level as all the other headers?
or maybe there is a more general solution to this via the main.css @vincentvis prob set this up?

frontend/src/views/GameView.vue
```diff
@@ -1,51 +1,167 @@
...
+			)
+				// this fills the above defined arc/circle
+				this.context.fill()
+			},
+				
+		// https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/fillRect
+		drawPaddles() {
+			let widthStart: number = 0;
+			const lineWidth: number = 15;
+			const lineHeight: number = this.heightPercentage(20);
```
livlaming, line 112: height percentage 12?

the paddles seem a little bit large/wide, what would you think of :

const lineHeight: number = this.heightPercentage(12);

frontend/src/views/GameView.vue
```diff
@@ -1,51 +1,167 @@
...
+
+			this.context.arc(
+					startingX,
+					startingY,
+					radius,
+					startAngle,
+					endAngle,
+			)
+				// this fills the above defined arc/circle
+				this.context.fill()
```
livlaming, line 105: perhaps the ball is a little large and could be a stroke?

this.context.lineWidth=4;
this.context.stroke()

frontend/src/views/GameView.vue
```diff
@@ -1,51 +1,167 @@
...
+					lineWidth,
+					lineHeight
+				);
+			},
+			
+		//  https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/arc
+		drawBall() {
+			const startingX: number = this.widthPercentage(40);
+			const startingY: number = this.heightPercentage(65)
+			const radius: number = 15;
```
livlaming, line 93: const radius: number = 12;
maybe just a tighny bit smaller?

livlaming (commented):
looks great!
some small comments on placements, widths, heights and strokes

frontend/src/views/GameView.vue
Ryno95: Changed to requested values

frontend/src/views/GameView.vue
Ryno95: very very nice style for the ball :D

PR #151, opened 2023-02-16
Li/account settings

### LINKED ISSUE #133: Create account-settings view
https://design.penpot.app/#/view/4755a563-5b43-8015-8001-f7f4126fea67?page-id=4755a563-5b43-8015-8001-f7f4126fea68&section=interactions&index=2

livlaming:
OMG, there will be a separate ticket for **uploading a new avatar**, but for the rest everything **besides 2fa** should work!

closes #133


frontend/src/views/SettingsView.vue
```diff
@@ -1,7 +1,116 @@
...
+				class="c_block c_form_group"
+				@submit.prevent="
+					userStore.updateProfile(user.id, updateProfileForm)
+				"
+			>
+				<InputField
+					v-if="this.user.intra_login"
+					label="intra name"
+					v-model="this.user.intra_login"
+					is_disabled="true"
```
Ryno95, line 21: should be:
``` typescript
:is_disabled="true"
```

Ryno95 (changes requested):
There is no check if the current_password actually matches the current password.

frontend/src/views/SettingsView.vue
vincentvis: otherwise it's a string literal value "true", and with the colon in front it becomes the js boolean value ;-)

frontend/src/stores/userStore.ts
```diff
@@ -12,13 +12,14 @@ export const useUserStore = defineStore('users', {
 		errors: [],
```
vincentvis, line 17: You can type this:
`errors: [] as String[]`

frontend/src/stores/userStore.ts
```diff
@@ -12,13 +12,14 @@ export const useUserStore = defineStore('users', {
...
+	getters: {
+		// getAllUsers: (state) => state.allUsers,
+		// getMe: (state) => state.me,
+	},
+	// actions == methods
 	actions: {
+
 		handleFormError(responseData: any) {
 			if (typeof responseData.message === 'string') {
 				this.errors = [responseData.message];
```
vincentvis, line 29: I found the better/recommended way for resetting an array:
```ts
this.errors.length = 0;
this.errors.push(responseData.message);
```

frontend/src/stores/userStore.ts
```diff
@@ -50,6 +51,28 @@ export const useUserStore = defineStore('users', {
...
+		async updateProfile(id: number, updateProfileForm: UpdateProfileForm) {
+			try {
+				if (updateProfileForm.new_password === '') {
+					delete updateProfileForm.new_password;
+					delete updateProfileForm.new_password_confirm;
+					delete updateProfileForm.password;
+				}
+				await patchRequest(`users/${id}`, updateProfileForm);
+				await this.refreshData();
+				await router.push('/settings');
```
vincentvis, line 66: is this needed? As in, do we always want to redirect to the settings page after a profile update? (it also seems to work without this line)

frontend/src/stores/userStore.ts
```diff
@@ -50,6 +51,28 @@ export const useUserStore = defineStore('users', {
...
+			} catch (e: any) {
+				if (typeof e.response.data.message === 'string') {
+					this.errors = [e.response.data.message];
```
vincentvis, line 69: Here the same as the other comment:

```ts
this.errors.length = 0;
this.errors.push(responseData.message);
```

frontend/src/views/ProfileView.vue
```diff
@@ -2,7 +2,7 @@
 	<div class="page_box_wrapper">
 		<div class="page_box">
 			<VerticalAvatarAndUserName
-				profile_picture="/test-profile.png"
+				:profile_picture="this.user.avatar"
```
vincentvis, line 5: inside of <template> thingies you don't need to reference "this"

frontend/src/views/ProfileView.vue
```diff
@@ -39,10 +39,12 @@ export default defineComponent({
 	async created() {
 		await useUserStore().refreshAllUsers();
 		const filteredUsers = useUserStore().allUsers.find((user: User) => Number(user.id) === Number(this.profile_id));
+
 		if (filteredUsers === undefined) {
 			this.$router.push('/profiles')
 		} else {
 			this.user = filteredUsers
+			console.log("avatar", this.user.avatar);
```
vincentvis: this can be removed

vincentvis: also, prettierjs for this file :P

vincentvis: also, should this also be updated to use the `storeToRefs` thing ? (maybe that's another issue)

frontend/src/views/SettingsView.vue
```diff
@@ -1,7 +1,110 @@
...
+import { useUserStore } from '@/stores/userStore';
+import { defineComponent, reactive } from 'vue';
+import type { UpdateProfileForm } from '@/types/User';
+import ProfileList from '@/views/ProfilesView.vue';
+import ProfileSettingAvatar from '@/components/user-info/ProfileSettingAvatar.vue';
+import ChangeAvatar from '@/components/input-fields/ChangeAvatar.vue';
+import { storeToRefs } from 'pinia';
+
+export default defineComponent({
+	name: 'RegisterView',
```
vincentvis, line 68: wrong component name

frontend/src/views/SettingsView.vue
```diff
@@ -1,7 +1,110 @@
...
+			<form
+				method="Post"
+				action=""
+				class="c_block c_form_group"
+				@submit.prevent="updateProfile(me.id, updateProfileForm)"
+			>
+				<InputField
+					v-if="me.is_intra"
+					label="intra name"
+					v-model="me.intra_login"
```
vincentvis, line 18: intra login does not exist on type User,
please add it in types/User.ts

frontend/src/views/SettingsView.vue
```diff
@@ -1,7 +1,110 @@
...
+					:is_disabled="true"
+				/>
+				<InputField
+					label="display_name"
+					:modelValue="me.name"
+					v-model="updateProfileForm.name"
```
vincentvis, line 24: vscode complains about this:
> An object literal cannot have multiple properties with the same name

i think it's about using :modelValue and v-model on the same component.. but not sure..

frontend/src/stores/userStore.ts
vincentvis:
```ts
this.errors.length = 0
```
after this would help hide previous errors

frontend/src/stores/userStore.ts
vincentvis: I also think that we need to figure something out that these are reset on every sending of the form, and at the change of a page.
Because now, if you do something "wrong" in the form, and then do the right thing, the errors don't disappear..

frontend/src/views/SettingsView.vue
vincentvis: if you fix your ESlint it should show it in clion as well :P

frontend/src/views/SettingsView.vue
```diff
@@ -1,7 +1,110 @@
...
+		const { updateProfile, refreshData } = userStore;
+		let updateProfileForm: UpdateProfileForm = reactive({
+			name: me.value.name,
+			new_password: '',
+			new_password_confirm: '',
+			password: '',
+		});
+		// if (me === undefined) {
+		// this.$router.push('/profiles');
+		// }
```
vincentvis, line 93: comment can be removed

frontend/src/views/SettingsView.vue
```diff
@@ -1,7 +1,110 @@
...
+		console.log(me.value.name);
```
vincentvis, line 94: console.log can be removed

vincentvis (changes requested):

frontend/src/components/user-info/ProfileSettingAvatar.vue
```diff
@@ -0,0 +1,33 @@
...
+<style scoped>
+.circle {
+	margin-left: auto;
+	margin-right: auto;
+	width: 10em;
+	height: 10em;
+	border: solid 2px black;
+	border-radius: 50%;
+	overflow: hidden;
+	margin-top: 1em;
```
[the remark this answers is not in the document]
livlaming: how would you translate this to this situation?

frontend/src/stores/userStore.ts
livlaming: this.errors.length = 0 ----->  = great!
and maybe it is nice to redirect to another page, to have a sense of the change?
if we want to also show our own profile in the profiles list we could redirect to that?
what would you think?

frontend/src/stores/userStore.ts
livlaming: I'm not sure, this is not goind wel if I copy paste this?

livlaming: I'm sure your solution is great but I cant make it work now

livlaming: something like this?

frontend/src/views/SettingsView.vue
livlaming: what about this?
:modelValue="me.name"
@update:modelValue="updateProfileForm.name = $event"

frontend/src/views/SettingsView.vue
livlaming: I wanted to keep this, as a reminder, what do we do is me doesnt existsbut we'll prob handle this with guards?

frontend/src/stores/userStore.ts
vincentvis: this should work:
```ts
{
   errors: [] as String[]
}
```

frontend/src/stores/userStore.ts
vincentvis: this is now great in the `handleFormError` function!
could you also quickly change the "updateProfile" to call that function in the catch block? (now it repeats the handleFromError logic)

frontend/src/views/SettingsView.vue
vincentvis: that should indeed be handled with guards at some point
if you wanna keep it, can you than add a textual part to the comment explaining that this should be implemented through a guard?
Because without that context the comment is odd

frontend/src/views/SettingsView.vue
vincentvis: it does not give an error 👍 
haven't ran the container, but assuming you actually filled in something and pressed save to verify that it works, it looks good to me :)

frontend/src/views/SettingsView.vue
livlaming: Ill create a ticket

frontend/src/stores/userStore.ts
livlaming: okay I'm redirecting to our own profile but we dont show our owns maybe something to discuss

frontend/src/views/ProfilesView.vue
```diff
@@ -28,8 +30,8 @@ export default defineComponent({
 	},
 	setup() {
 		const userStore = useUserStore();
-		const { refreshData, allUsers, isBlocked, getCurrentRel } = userStore;
-
+		const { refreshData, isBlocked, getCurrentRel } = userStore;
+		const {me , allUsers} = storeToRefs(userStore)
```
vincentvis, line 34: this didn't go through prettierjs

vincentvis (approved):

frontend/src/views/ProfileView.vue
livlaming: haah thingies

Ryno95 (approved):

PR #156, opened 2023-02-17
Ry/game movement

### LINKED ISSUE #68: Implement movement with button pushes
Have button press and release passed to backend where the position of elements will be handled.
Think of having types (interface for vue, class for nest) that are the same on both sides

Ryno95:
Basic movement implemented. No scoring or game end logic yet, thats for the next ticket. Also I still need to implement a random starting point for the ball.

Added lots of hopefully clear comments in the code itself, so not going to deep into things here.

You can us "w" and "s" to move player1 paddle (left).

You can use arrowUp and arrowDown to move player2 paddle (right)

Used [this tutorial](https://developer.mozilla.org/en-US/docs/Games/Tutorials/2D_Breakout_game_pure_JavaScript) as inspiration for movement and collisions:
closes #68


backend/src/pong/pong.gateway.ts
```diff
@@ -0,0 +1,51 @@
...
+	WebSocketServer,
+} from '@nestjs/websockets';
+import { Server } from 'socket.io';
+import { GameState, MovementKeys } from './Game.types';
+import { PongService } from './pong.service';
+
+@WebSocketGateway({
+	namespace: '/pong',
+	cors: {
+		origin: '*:*',
```
vincentvis, line 15: `'*'`?

backend/src/pong/pong.service.ts
```diff
@@ -0,0 +1,155 @@
...
+
+		//  check collision for paddle one
+		if (this.doesHitPaddle(gameState.ball, paddleP1, gameState.canvas)) {
+			console.log('Collision Paddle one');
+			const paddleCenter = paddleP1.position.y + paddleP1.height / 2;
+			const d = paddleCenter - ballPos.y;
+			//  speed up the ball
+			this.dx = -(this.dx - 0.02);
+			// change bounce angle
+			this.dy -= d * -0.01;
```
vincentvis, line 96: does the ball increasingly become a little bit faster without limit?
and same-ish question about the angle

i haven't played it, can also be fine since the values are quite low, just curious how it'll play

backend/src/pong/pong.service.ts
```diff
@@ -0,0 +1,155 @@
...
+			pos + direction > fullHeigthOrWidth - radius || pos + direction < radius
+		);
+	}
+
+	doesHitPaddle(ball: Ball, rect: Paddle, canvas: Position) {
+		const paddleY = rect.position.y;
+		const paddleHeight = rect.height;
+		const paddleTop = paddleY - paddleHeight / 2;
+		const paddleBottom = paddleY + paddleHeight / 2;
+		const magicOffset = 0.2;
```
vincentvis, line 62: it sounds like `magicOffset` is related to the ball radius, or is it actual magic somehow?
And if it is related to ball radius, should we use that to calculate it?

vincentvis (approved):

backend/src/pong/pong.service.ts
Ryno95: The speed does, the angle doesn't. "d" value for the changing of the y-angle ranges from +-"0.01 - 0.7" but changes for every impact.

As for the speed, I can limit it, but it increases very slowly, so by the time it's quite fast the point has been going for some time already.

backend/src/pong/pong.service.ts
Ryno95: changed it to actually make sense haha
`magicOffset => ballLineThicknessOffset`
This is for the visual aspect of the collision, it seems the thickness of the line is not part of the size of the circle.

backend/src/pong/Game.types.ts
```diff
@@ -0,0 +1,30 @@
+export class Position {
+	x: number;
+	y: number;
+}
+
```
livlaming, line 5: are these types specially for the backend otherwise why are there two games.ts files?

backend/src/pong/pong.service.ts
```diff
@@ -0,0 +1,156 @@
...
+		const fullHeigthOrWidth = 100;
+		return (
+			pos + direction > fullHeigthOrWidth - radius || pos + direction < radius
+		);
+	}
+
+	doesHitPaddle(ball: Ball, rect: Paddle, canvas: Position) {
+		const paddleY = rect.position.y;
+		const paddleHeight = rect.height;
+		const paddleTop = paddleY - paddleHeight / 2;
```
livlaming, line 60: is this paddle centre than? because of the /2?

backend/src/pong/pong.service.ts
```diff
@@ -0,0 +1,156 @@
...
+		const paddleBottom = paddleY + paddleHeight / 2;
+		const ballLineThicknessOffset = 0.2;
+		const paddleWall =
+			ball.radius - ballLineThicknessOffset + this.paddleWidth / 2;
+
```
livlaming, line 65: would be lovely to also fill the ball buts that prob in a different file!

backend/src/pong/pong.service.ts
```diff
@@ -0,0 +1,156 @@
...
+		// if ball will hit the invisible wall that extends from the paddle
+		// and ball position is within the paddle top and bottom
+		// console.log(paddleWall);
+		if (
+			this.doesHitWall(paddleWall, ball.position.x, this.dx) &&
```
livlaming, line 70: I do have the feeling that the balls goes a little bit through the walls top and bottom for the rest it looks nicE!

backend/src/pong/pong.gateway.ts
```diff
@@ -0,0 +1,51 @@
...
+})
+export class PongGateway implements OnGatewayConnection {
+	constructor(private readonly pongService: PongService) {}
+	@WebSocketServer()
+	server: Server;
+
+	private gameState: GameState;
+
+	async handleConnection() {
+		console.log('\n!Socket is connected!\n');
```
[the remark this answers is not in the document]
Ryno95: Socket connect is nice to have to see when the connection happens, I'm removing the logging if the game state

backend/src/pong/Game.types.ts
Ryno95: Is needed as the type files are in two different containers... we can maybe do a symlink, not sure if this copies it to the both docker files though....? Do you know?

backend/src/pong/pong.service.ts
Ryno95: black ball then?

backend/src/pong/Game.types.ts
Ryno95: For now renamed them with either a .fe or .be

backend/src/pong/pong.service.ts
Ryno95: Nvrmind, black lined ball with white filling, #OreoBall

backend/src/pong/Game.types.ts
livlaming: cool

backend/src/pong/pong.service.ts
livlaming: not sure where you start counting with y but it sounds like the exact opposite haha but if this works its fine

backend/src/pong/pong.service.ts
livlaming: perfect!

backend/src/pong/pong.service.ts
livlaming: cool

backend/src/pong/pong.service.ts
Ryno95: count starts on top left corner, x = 0, y = 0, so right bottom corner x = width and y = height.
So the lower y becomes for a element the higher up it'll be on the canvas :)
It works exactly opposite to how we are used to counting cause y - 1 is up, and y + 1 is down.

backend/src/pong/pong.service.ts
vincentvis: If the speed does not increase at an unreasonable rate I'm fine with it :P It should probably at some point become too fast so we "force" a point to make the games not infinite

backend/src/pong/pong.gateway.ts
```diff
@@ -6,13 +6,13 @@ import {
 	WebSocketServer,
 } from '@nestjs/websockets';
 import { Server } from 'socket.io';
-import { GameState, MovementKeys } from './Game.types';
+import { GameState, MovementKeys } from './game.types.be';
```
vincentvis, line 9: ".be"? is this the Belgian version?

backend/src/pong/pong.service.ts
```diff
@@ -117,39 +124,33 @@ export class PongService {
...
 
+	moveUp(paddlePosY: number) {
+		const topMax = this.paddleHeight / 2;
+		return Math.max(paddlePosY - 1, topMax);
+	}
+
+	moveDown(paddlePosY: number) {
+		const bottomMax = 100 - this.paddleHeight / 2;
+		return Math.min(paddlePosY + 1, bottomMax);
+	}
```
vincentvis, line 135: NICE! lekker
goeie refactor :)

vincentvis (approved):

PR #157, opened 2023-02-20
Vi/chat socket clean

### LINKED ISSUE #146: Create AppSocket service
For the data flow, this is now my idea:

```
on start of app, socket connection is opened which listens to "update" events
       ┃ 
┏━━━━━━┻-> socket receives update event and updates the store with new info (and thus view is updated)
┃
┃
┃    (1) client sends post request to update data
┃         ┃
┃         ┗-> backend recieves post request and handles the service calls
┃                ┃
┃                ┗-> backend emits socket message with update data
┗━━━━━━━━━━━━━━━━━━━━━━┫
                       ┗-> backend returns OK result to client
  
```

I will create a AppSocket for this so that all communication can go through one socket
Maybe we want to split this in a per-store socket? Not sure.. 
Idea is now that there will be an AppStore socket which can handle update messages for different types of data and then updates that data in the correct store. 

This replaces the "`refreshData()`" call to the stores after something is updated. 
Technically we only have to "get" a thing once, when that thing is needed. And afterwards when there are updates to that data they are done through the socket.

So if /user/25 is loaded:
- you check userStore if "id=25" exists in the available data
  - if not
     - do a get request to get the data
     - show the profile
  - if yes
     - show the profile

if user with id=25 is updated, the backend transmits a "user_update" message with the data ("user_update" is an example name!)
the frontend socket receives this message, checks if user with id=25 exists and then updates the user store with the new data.

vincentvis:
closes #146


backend/src/chats/chat/chat.controller.ts
```diff
@@ -39,7 +54,53 @@ export class ChatController {
...
+							avatar: user.avatar,
+						})) ?? [],
+				},
+			};
+			this.socketService.chatlist_emit('all', socketMessage);
+		}
+		return chat;
 	}
 
 	@Get()
```
livlaming, line 106: ?

backend/src/chats/chat/chat.controller.ts
```diff
@@ -59,6 +120,13 @@ export class ChatController {
 		});
 	}
 
+	@Get(':id/messages')
+	async chatMessages(@Param('id') id: number) {
```
livlaming, line 124: is this something we use? or maybe if we join a channel later, and we would like to see all the current messages? and prob also with general channels, in which we can only look at the messages?

backend/src/chats/chat/dto/create-chat.dto.ts
```diff
@@ -12,4 +13,7 @@ export class CreateChatDto {
 
 	@IsOptional()
 	type?: string;
+
+	@IsOptional()
```
livlaming, line 17: should this always have users? as in not optional?

backend/src/chats/chat/entities/chat.entity.ts
```diff
@@ -52,7 +52,7 @@ export class Chat {
 	@Exclude()
 	password: string;
 
-	@OneToMany(() => Message, (message) => message.chat_id)
+	@OneToMany(() => Message, (message) => message.chat)
 	messages: Message[];
```
livlaming, line 56: what would be the difference between chat_id or chat? or is it just different naming?

backend/src/chats/chats.gateway.ts
```diff
@@ -0,0 +1,51 @@
...
+	WebSocketServer,
+	ConnectedSocket,
+} from '@nestjs/websockets';
+import { Server, Socket } from 'socket.io';
+import { SocketService } from '../socket/socket.service';
+import { ChatId } from '../socket/socket.types';
+
+@WebSocketGateway({
+	namespace: '/chats',
+	cors: {
```
livlaming, line 17: Cross-Origin Resource Sharing is an http-header?????

backend/src/chats/chats.gateway.ts
```diff
@@ -0,0 +1,51 @@
...
+		origin: '*',
+	},
```
livlaming, line 19: whats does the start do?

backend/src/chats/message/entities/message.entity.ts
```diff
@@ -25,13 +26,19 @@ export class Message {
...
 
-	@ManyToOne((type) => Chat, (chat) => chat.id)
-	@JoinColumn({ name: 'chat_id' })
-	chat_id: Chat;
+	@RelationId((message: Message) => message.sender_id)
+	user_id: number;
+
+	@ManyToOne(() => Chat, (chat) => chat.id, { nullable: false })
+	@JoinColumn({ name: 'chat' })
+	chat: Chat;
```
livlaming, line 38: is this a chat_id? probably nottt?

backend/src/chats/message/message.controller.ts
```diff
@@ -9,35 +9,80 @@ import {
...
+				},
+			};
+			this.socketService.chatServer.emit('messageListUpdate', socketMessage);
+		}
+		return newMessage;
 	}
 
+	// @NOTE
+	// maybe we also should not have this..
+	// i think it is never used anyway, and this makes it so that any logged in user can always get _all_ messages (also of other people..)
```
livlaming, line 50: agreed!

backend/src/chats/message/message.controller.ts
```diff
@@ -9,35 +9,80 @@ import {
...
 	@Get()
 	findAll() {
 		return this.messageService.findAll({
-			relations: { chat_id: true, sender_id: true },
+			relations: { chat: true, sender_id: true },
 		});
 	}
 
 	@Get(':id')
```
livlaming, line 58: when would you need this?

backend/src/seeders/seed.data.ts
```diff
@@ -26,7 +26,7 @@ class seedGame {
 
 class seedMessage {
 	sender_id: number;
-	chat_id: number;
+	chat: number;
```
livlaming, line 29: ah cool so this is the chat id? same qyestion as earlier.. must say that I liked the id within the chat_id but maybe its meaning changed?

backend/src/socket/socket.service.ts
```diff
@@ -0,0 +1,157 @@
...
+		);
+		try {
+			const userId: number = this.authService.userIdFromCookieString(cookie);
+			await this.userService.findOne({ where: { id: userId } });
+			const roomName = this.chatIdToRoomName(chatId);
+			if (!this.messageLists[chatId]) {
+				this.messageLists[chatId] = {
+					chatId: chatId,
+					roomName: roomName,
+					subscribers: [],
```
livlaming, line 43: does this one need a type?

livlaming (commented):
allrrrighttttt looks great,
v large, so I'm not sure if I catched everything butttttt
 one general thing the I already mentioned, but maybe it's nice to only mention the person you're chatting to in the dm list

backend/src/chats/chat/chat.controller.ts
vincentvis: with this comment you made me look in the file. Scan over the relevant lines. Try to figure out what you mean by the singular question mark. And then still be frustrated since I can't figure out what you mean or why you commented here....
So a little more specific please ;)

backend/src/chats/chat/chat.controller.ts
vincentvis: 1. yes it is used, it gets the messages from a specific chat. you can find it in `getActiveChatMessages` action from `messageStore`
2. it doesn't matter when you join a channel, it's just for getting the messages from any chat to the frontend
3. not sure what you mean with the last question :)

backend/src/chats/chat/dto/create-chat.dto.ts
vincentvis: that would probably be good :P 
but since I kind of half-assed it while adding this functionality that would be a different ticket (made a ticket #162)

backend/src/chats/chat/entities/chat.entity.ts
vincentvis: I renamed the relationship in messages from `chat_id` to `chat` and added a `chat_id` column that actually stores the relationid (see message.entity)

backend/src/chats/chats.gateway.ts
vincentvis: yes

frontend/src/stores/socketStore.ts
```diff
@@ -0,0 +1,63 @@
...
+			);
+			this.chats.socket.on(
+				'messageListUpdate',
+				(update: SocketMessage<SingleMessage>) => {
+					useMessageStore().socketAction(update);
+				},
+			);
+		},
+		disconnect() {
+			this.chats.socket?.off('newMessage');
```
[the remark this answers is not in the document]
vincentvis: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

> The optional chaining (?.) operator accesses an object's property or calls a function. If the object accessed or function called using this operator is undefined or null, the expression short circuits and evaluates to undefined instead of throwing an error.

frontend/src/stores/userStore.ts
```diff
@@ -18,14 +18,16 @@ export const useUserStore = defineStore('users', {
...
 
 	// getters == computed values
-	getters: {
-	},
+	getters: {},
 	// actions == methods
 	actions: {
+		// this should be moved out of the userStore
+		// it is not userStore functionality
+		// and it should be typed with something other than any..
```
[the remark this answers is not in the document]
vincentvis: yes, should probably make a ticket out of this.. (created a draft ticket for now)

frontend/src/types/Sockets.ts
```diff
@@ -0,0 +1,18 @@
+import type { Socket } from 'socket.io-client';
+
+export interface SocketChat {
+	socket: Socket | null;
```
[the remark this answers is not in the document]
vincentvis: it defines the type of the `socket` property of the SocketChat interface.
`|` == OR
so in this case the property socket can be of type `Socket` OR of type `null`

https://www.typescriptlang.org/docs/handbook/2/objects.html

frontend/src/types/Chat.ts
```diff
@@ -17,13 +17,21 @@ export interface UserChatPermissions {
 export interface SingleMessage {
 	id: number;
 
-	sender: User;
-	chat: Chat;
+	sender?: User;
```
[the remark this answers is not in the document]
vincentvis: sender would be the full relationship object of the user. There is a `user_id` property that is used for when this object is used for communicating through sockets.

the sender property is mainly there as a fallback for when user_id is not set while rendering the message component (actually made a change there because of your comment here :P)

frontend/src/types/Chat.ts
```diff
@@ -17,13 +17,21 @@ export interface UserChatPermissions {
...
+	chat?: Chat;
```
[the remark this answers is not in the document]
vincentvis: I removed the chat property, only `chat_id` is used

backend/src/chats/message/entities/message.entity.ts
vincentvis: is what a chat_id? this column name? I renamed it to 'chat', since it's a full relation, and when loaded, resolves to a full chat object. And added a chat_id property (just below this one) with the @RelationId decorator to save the ID

backend/src/chats/message/message.controller.ts
```diff
@@ -9,35 +9,80 @@ import {
...
 			where: { id },
-			relations: { chat_id: true, sender_id: true },
+			relations: { chat: true, sender_id: true },
 		});
 	}
 
 	@Delete(':id')
-	remove(@Param('id') id: number) {
-		return this.messageService.remove(id);
+	async remove(@Param('id') id: number) {
```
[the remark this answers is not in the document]
vincentvis: should probably never happen, was just handy for testing. we can make an issue to remove this before handing it in

backend/src/me/me.module.ts
```diff
@@ -1,22 +1,13 @@
...
 @Module({
-	imports: [
-		SharedModule,
-		AuthModule,
-		UserModule,
-		ChatModule,
-		MessageModule,
-		ChatUserPermissionModule,
-	],
+	imports: [UserModule, ChatModule, MessageModule, ChatUserPermissionModule],
```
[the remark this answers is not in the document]
vincentvis: aside from that we now have a pre-commit hook that should do prettier anyway, my prettier is also running fine in vscode, so this file has gone through prettierjs, yes.

backend/src/seeders/seed.data.ts
vincentvis: the name of the relationship changed so it also changed here

backend/src/socket/socket.service.ts
vincentvis: subscribers is typed,
in socket.types.ts you can find the definition for MessageList type and ChatInfo interface
```ts
export type UserId = number;

export interface ChatInfo {
	chatId: ChatId;
	roomName: string;
	subscribers: UserId[];
}

export type MessageList = Record<ChatId, ChatInfo>;
```

backend/src/socket/socket.types.ts
```diff
@@ -0,0 +1,44 @@
...
+
+export type Chat_Type = 'dm' | 'channel';
+export type ActionType = 'new' | 'update' | 'delete';
+
+export interface SingleMessage {
+	id: number;
+
+	user_id?: number;
+	chat_id: number;
+	content?: string;
```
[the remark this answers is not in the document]
vincentvis: the interface is also used without sending content, for now at least, in the message delete method

frontend/src/stores/chatStore.ts
```diff
@@ -1,37 +1,124 @@
...
+		subscribed: false,
+		initialized: false,
 	}),
-  // getters == computed values
+	// getters == computed values
 	getters: {
-		getMyChats: () => useUserStore().getMe.chats,
-        getAllChats: (state) => state.allChats,
+		getAllChats: (state) =>
+			[...state.dms, ...state.channels].sort((a, b) => a.id - b.id),
```
Ryno95, line 17: the above can be typed liked this as well right? 
`[...state.dms, ...state.channels].sort((dm, channel) => dm.id - channel.id)`
so sort param 1 = a dm, param 2 = a channel?
then sorting on `dm.id - channel.d`
Could you maybe elaborate on what's happening here?

frontend/src/stores/userStore.ts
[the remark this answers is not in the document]
Ryno95: This specifically handles the errors array for the userStore right? Should we mak it a util that returns an array of errors?

Ryno95 (commented):
A short description in the PR would be nice, gives us an idea of what the functionality is of what we are looking at. 

It would be nice if a we get a notification when something changes in the database so we know to prune docker and remove the .db file.

Pushed some example comments in chat.controller that might be chilled to add when logic becomes a bit cluttered and hard to interpret with a casual view. 

The Socket part is super clear, very nice!

backend/src/chats/chat/chat.controller.ts
livlaming: haha I think I saw something empty?

backend/src/chats/chat/chat.controller.ts
livlaming: cool

backend/src/chats/message/message.controller.ts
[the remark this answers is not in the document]
livlaming: cool!

frontend/src/stores/userStore.ts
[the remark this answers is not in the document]
vincentvis: but don't we have forms on other pages as well?
We could make a "formStore"? not sure about this anymore...

backend/src/chats/message/message.controller.ts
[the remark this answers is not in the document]
vincentvis: correct :)

backend/src/socket/socket.service.ts
```diff
@@ -0,0 +1,157 @@
...
+		return `room_${chatId}`;
+	}
+
+	async joinRoom(chatId: ChatId, socket: Socket) {
+		const cookie: string = this.jwtCookieFromHandshakeString(
+			socket.handshake.headers.cookie,
+		);
+		try {
+			const userId: number = this.authService.userIdFromCookieString(cookie);
+			await this.userService.findOne({ where: { id: userId } });
```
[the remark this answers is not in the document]
vincentvis: correct. It's quite hard to fake a user ID in a valid JWT token (since it uses a secret from our .env), so as an "attack vector" it's not really an issue here. But if for example a user would be deleted and then comes back this would catch that user not existing any more.

frontend/src/stores/chatStore.ts
vincentvis: there are 2 arrays in the component. DM's and Channels. 
the first part just puts them both in 1 array:
```ts
[...state.dms, ...state.channels]
```
the order in the new array at that point is the dms first, then the channels, which is not necessarily the order of their ID's (dms can be created later than channels etc)
so it's sorted on ID, because afterwards finding something is faster if the data is sorted.

the sorting happens on the joined array of dms and channels, where they have kinda lost their meaning of if they are a dm/channel, since they are now  both just in a list of "chats" (their type becomes irrelevant)

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort

`.sort()` is an Array method where (a,b) are the usual variable names. Just because it compares a to b. Javascript abstracts the sorting algorithm behind this, so you only supply a function to determine if `a` should be placed after `b`, or if `a` should be placed before `b` (the docs page is quite good... better than what I'm trying here..)

maybe I'm over explaining... I'll stop.. ask me if it's unclear :P

Ryno95: hehe knowing that `[...state.dms, ...state.channels]` merges two array, it's clear thanks! :D

frontend/src/stores/userStore.ts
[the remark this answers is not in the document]
Ryno95: We can maybe make it a helper function that returns the array of errors? This way it can be used everywhere but the errors are kept wherever you want to use them?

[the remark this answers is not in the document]
Ryno95: Feel free to resolve this after reading this

backend/src/socket/socket.service.ts
```diff
@@ -0,0 +1,157 @@
...
+			}
+		} else if (to instanceof Array && to.length > 0) {
+			to.forEach((userId: number) => {
+				if (this.chatListSubscribers[userId]) {
+					this.chatListSubscribers[userId].forEach((socket) => {
+						socket.emit('chatListUpdate', message);
+					});
+				}
+			});
+		}
```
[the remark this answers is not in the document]
Ryno95: Bon!

livlaming (approved):

PR #159, opened 2023-02-21
Ry/endpoint protection and refactor

### LINKED ISSUE #87: Protect endpoints with guards
- [x] determine which endpoints should only be accessible as authenticated user
- [ ] determine which endpoints should not be accessible as authenticated user
- [ ] determine which endpoints should be removed (/are not needed)

Ryno95:
Added AuthGuard to all controllers.  

Removed Achievement controller and dto's as it is only used for seeding. 

More controllers can be cleaned up, commented out some possible endpoints to remove in user-achievement.

closes #87


livlaming (approved):
things are working maybe a v short chat about the resoluation or the conceptual outcome of your journey but for not do merge if @vincentvis aggreeessss

backend/src/auth/auth.module.ts
```diff
@@ -2,10 +2,9 @@ import { forwardRef, Module } from '@nestjs/common';
 import { AuthController } from './auth.controller';
 import { AuthService } from './auth.service';
 import { UserModule } from '../users/user/user.module';
-import { SharedModule } from '../shared/shared.module';
 
 @Module({
-	imports: [forwardRef(() => UserModule), SharedModule],
+	imports: [forwardRef(() => UserModule)],
```
vincentvis, line 7: Did you change something else?
Because with me it does not compile if i remove the `SharedModule`..

backend/src/chats/chat/chat.module.ts
```diff
@@ -1,11 +1,17 @@
...
 import { TypeOrmModule } from '@nestjs/typeorm';
 import { Chat } from './entities/chat.entity';
+import { AuthModule } from '../../auth/auth.module';
+import { SharedModule } from '../../shared/shared.module';
 
 @Module({
-	imports: [TypeOrmModule.forFeature([Chat])],
+	imports: [
+		TypeOrmModule.forFeature([Chat]),
+		forwardRef(() => AuthModule),
```
vincentvis, line 12: I don't seem to need to forwardRef here

vincentvis: Actually.. It seems to work without the import of the AuthModule and the Shared module.. Maybe those two are only needed when you actually do something withthe auth controller or service. Not just the guard.. (not sure..)

backend/src/chats/message/message.module.ts
```diff
@@ -1,11 +1,17 @@
...
 import { Message } from './entities/message.entity';
+import { AuthModule } from '../../auth/auth.module';
+import { SharedModule } from '../../shared/shared.module';
 
 @Module({
-	imports: [TypeOrmModule.forFeature([Message])],
+	imports: [
+		TypeOrmModule.forFeature([Message]),
+		forwardRef(() => AuthModule),
+		SharedModule,
```
vincentvis, line 13: so here I only need the AuthModule (without forwardRef), not the SharedModule... (no idea how this works)

backend/src/pong/game/game.module.ts
```diff
@@ -1,11 +1,17 @@
...
 import { Game } from './entities/game.entity';
+import { AuthModule } from '../../auth/auth.module';
+import { SharedModule } from '../../shared/shared.module';
 
 @Module({
-	imports: [TypeOrmModule.forFeature([Game])],
+	imports: [
+		TypeOrmModule.forFeature([Game]),
+		forwardRef(() => AuthModule),
+		SharedModule,
```
vincentvis, line 13: don't need either of them..

backend/src/pong/game_invite/game-invite.module.ts
```diff
@@ -1,11 +1,17 @@
...
 import { GameInvite } from './entities/game-invite.entity';
+import { AuthModule } from '../../auth/auth.module';
+import { SharedModule } from '../../shared/shared.module';
 
 @Module({
-	imports: [TypeOrmModule.forFeature([GameInvite])],
+	imports: [
+		TypeOrmModule.forFeature([GameInvite]),
+		forwardRef(() => AuthModule),
+		SharedModule,
```
vincentvis, line 13: don't seem to need them

backend/src/shared/abstract.service.ts
```diff
@@ -1,4 +1,4 @@
-import { Injectable, NotFoundException } from '@nestjs/common';
+import { Injectable, NotFoundException, SetMetadata } from '@nestjs/common';
```
vincentvis, line 1: unused import

backend/src/users/user/user.controller.ts
```diff
@@ -95,7 +94,6 @@ export class UserController {
 		});
 	}
 
-	@UseGuards(AuthGuard)
```
vincentvis, line 98: This probably still needs the guard right? Or do we want to be able to update a user without being authenticated?

vincentvis (changes requested):
38 files.. lol
but I guess most of my comments are the same, so my 13 comments are more 2 or something like that..

so if you can see if it also goes fine on your side without those imports that is my change request
if it doesn't work let me know..

backend/src/auth/auth.module.ts
Ryno95: I'm assuming you didn't compile this branch but another one?🧐

backend/src/chats/chat/chat.module.ts
Ryno95: Me copy and paste without double checking necessity, never ;)

backend/src/chats/message/message.module.ts
Ryno95: I can remove this for this branch as well, but do you need it in your chat branch? Then I'll leave it in...

backend/src/users/user/user.controller.ts
Ryno95: Yes thank you miss this one!

Ryno95:
Soooo i think I started adding this everywhere after the global APP_GAURD was too much of a hassle, then thought, "There has to be an easier way", there  was but I forgot to remove the includes everywhere hehe thanks V!

PR #160, opened 2023-02-21
added Vue routing guard

### LINKED ISSUE #119: Protect pages with navigation guard
https://router.vuejs.org/guide/advanced/navigation-guards.html#navigation-guards

Ryno95:
Guard implemented, it only works with local login, seems the intra login is not working fully yet on the FE. I made a separate issue ticket for that.

Also as soon as the logout view is working I have to check that this works as well.
Edit. Pulled Lisa's branch in here to test, works if you set this.isLoggedIn to false;

closes #119


backend/src/auth/auth.service.ts
```diff
@@ -115,7 +115,7 @@ export class AuthService {
 
 	async login(id: number, response: Response) {
 		const jwt = await this.jwtService.signAsync({ id: id });
-		response.cookie('jwt', jwt, { httpOnly: true });
+		response.cookie('jwt', jwt, { httpOnly: false });
```
vincentvis, line 118: I think setting this to httpOnly is considered bad practice.. something with security
but since you're using it on the frontend I'm not sure how to resolve this..

maybe we can use the userStore.login method to save some boolean if the post call there was successful. And just use that to keep track of if people are logged in?  (i see a boolean there already, so not 100% what you tried already)

Ryno95: Yes I tried that for a bit, but every time the page is refreshed the boolean is set to false, then tried adding it to the userstore.me, but the friggen async getMe requests fucks me, hard.
Tried this way, it worked, and left it at that hehe

vincentvis (approved):

PR #161, opened 2023-02-21
Li/logout

livlaming:
greetings from ryno


frontend/src/App.vue
```diff
@@ -41,7 +42,7 @@ import { RouterLink, RouterView } from 'vue-router';
 				<div class="nav_ball"></div>
 				Profiles
 			</RouterLink>
-			<RouterLink to="/logout">
+			<RouterLink v-on:click='useUserStore().logout()' to="#">
```
vincentvis, line 45: i think this should become just an `<a>` element
or a `<routerlink to="{name:login}">` with an `beforeLeave` hook that calls logout?

I think Ryno as also working on something with `beforeLeave` so maybe you can look there for some inspiration/example

PR #167, opened 2023-02-25
Ry/relationships with sockets

Ryno95:
Built a RelationshipStore instead of doing everything via the userStore.

Changed the userRelationship entity so remember to delete your .db file, prune backend img, and restart the docker network!

Added the specifierId column in the UR.entity and DTO's. This way we can only unblock a user if the specifierId ==== me.id.

Added a new endpoint in me.controller for relationships, only sending the needed data now.

Seems like a lot of files, a lot of prettier edits.... finally hehe

When relationship is updated via the frontend, it emits to the backend which then emits to all involved users that an update has happened an that the rels need to bee refreshed.


backend/src/users/user-relationship/entities/user-relationship.entity.ts
```diff
@@ -15,25 +15,25 @@ export enum validRelationships {
...
 	})
-	@JoinColumn({ name: 'target_id' })
-	target_id: User;
+	@JoinColumn({ name: 'target' })
+	target: User;
 
-	// @Column()
-	// specifier_id: number;
+	@Column()
+	specifier_id: number;
```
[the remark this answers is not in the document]
vincentvis: soooooooo I figured out that it is used and how

If A blocks B
a relationship between A and B is made (A=source, B=target)
if A unblocks B, the relationship is updated to type `none`
if B blocks A, the same relationship is updated to `blocked` again (and specifier_id is set) The specifier ID is then used to determine who blocked who, and not the source, since it's the same relationship reused.

i'm now thinking that it could have been done by removing the old relationship when you unblock. But this works, so alll goood :))

backend/src/users/user-relationship/entities/user-relationship.entity.ts
```diff
@@ -15,25 +15,25 @@ export enum validRelationships {
 }
 
 @Entity('user_relationships')
-@Check(`"source_id" <> "target_id"`) // 				makes [2,2] impossible
+@Check(`"source" <> "target"`) // 				makes [2,2] impossible
```
vincentvis, line 18: we can have multiple rows with the same source/target
which causes some problems
to partially solve this we can add:

```ts
@Unique(['source','target'])
```
which solves `[{2,3},{2,3}]` but not `[{2,3},{3,2}]`

backend/src/users/user-relationship/user-relationship.gateway.ts
```diff
@@ -0,0 +1,105 @@
...
+	@WebSocketServer()
+	server: Server;
+	public relationshipObservers: RelationshipObservers = {};
+
+	jwtCookieFromHandshakeString(string: string) {
+		return string
+			.split(' ')
+			.find((cookie) => cookie.startsWith('jwt='))
+			?.slice(4);
+	}
```
vincentvis, line 44: you call this from the socketService

backend/src/users/user-relationship/user-relationship.gateway.ts
```diff
@@ -0,0 +1,105 @@
...
+
+	handleDisconnect(client: Socket) {
+		const cookie: string = this.jwtCookieFromHandshakeString(
+			client.handshake.headers.cookie,
+		);
+		const userId: number = this.authService.userIdFromCookieString(cookie);
+		this.relationshipObservers[userId].rooms.forEach((roomId) => {
+			client.leave(roomId);
+		});
+		this.relationshipObservers[userId].rooms = [];
```
vincentvis, line 95: this should have a check around it to check if `userId` exists in Observers

backend/src/users/user-relationship/user-relationship.gateway.ts
```diff
@@ -0,0 +1,105 @@
...
+		} catch (e) {
+			console.log();
+		}
+	}
+
+	handleDisconnect(client: Socket) {
+		const cookie: string = this.jwtCookieFromHandshakeString(
+			client.handshake.headers.cookie,
+		);
+		const userId: number = this.authService.userIdFromCookieString(cookie);
```
vincentvis, line 91: this should be in a try/catch block since the auth.service throws if the jwt is invalid

backend/src/users/user-relationship/entities/user-relationship.entity.ts
vincentvis: *kuch*

nevermind

frontend/src/stores/relationshipStore.ts
```diff
@@ -0,0 +1,164 @@
...
+	// getters == computed values
+	getters: {},
+	// actions == methods
+	actions: {
+		async initialize() {
+			if (this.isInitialized === false) {
+				// for when sockets are setup
+				this.isInitialized = false;
+				this.socket = io('http://localhost:8080/user/relationship');
+				await this.refreshRelationships();
```
vincentvis, line 24: not sure, but should this be before init of socket? probably doesn't matter that much...

vincentvis (changes requested):

frontend/src/views/ProfilesView.vue
```diff
@@ -17,55 +17,35 @@
...
 	setup() {
+		const relationshipStore = useRelationshipStore();
+		const { getSingleRelationship } = relationshipStore;
+		relationshipStore.initialize();
+
 		const userStore = useUserStore();
-		const { refreshData, isBlocked, getCurrentRel } = userStore;
+		const { refreshData } = userStore;
 		const { me, allUsers } = storeToRefs(userStore);
+		refreshData();
```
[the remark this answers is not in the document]
Ryno95: This is the refresh for all users, but this should be initiated on login so did that and removed the refresh here :)

backend/src/users/user-relationship/entities/user-relationship.entity.ts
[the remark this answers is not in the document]
Ryno95: exactly, as soon as a rel exists, it is only ever updated. indeed via delete and then create is just as viable

backend/src/users/user-relationship/user-relationship.gateway.ts
Ryno95: This is a statement hehe
I copied it from socketService, I didn't want to instantiate a socketService for only this purpose, lazy I know.
Added it to socket.utils.ts and replaced usage everywhere

vincentvis (approved):
supertrouper

PR #171, opened 2023-02-26
Change chat_list_items profile pictures

### LINKED ISSUE #170: Chat_list_item profile pictures improvement
fix that profile picture array can be empty

vincentvis:
- adds temporary default avatar (should probably be something else)
- changes the sorting order of the visible avatars and adds a fallback avatar (so there is never none)

closes #170


Ryno95 (approved):
Very sexy default ;)

PR #173, opened 2023-02-27
quick patch userStore form error

vincentvis:
message can be undefined if the backend server is failing at somethin…g else than processing the form


Ryno95 (approved):
Nois!

PR #174, opened 2023-02-27
Vi/chat join

### LINKED ISSUE #168: Chat frontend - when user views channel but is not a member yet, show a "join channel" button instead of "send message"
- should set 'read' and 'post' permissions for current user in the chat
- should emit a socket message to everybody in the chat that user has joined

vincentvis:
User that is not in a chat, can now join a public chat by pressing a button

adds logic to frontend to show button or "send message" part
adds logic to backend to add a member and emit a message

closes #168


Ryno95 (commented):
Not sure what this is doing.....

I tried registering a user to join one of the chats but that retunrs:
image

I tried to join a chat, but either a user is already in all visible channels or no chats are visible?

I made a new public channel with one user in there, but i cant seem to join?

Most probably I'm missing something but I can't really test or see any new functionality....

Ryno95 (approved):
The join functionality works like a dream. 

Only thing was the private chats, I think you fixed it already.

PR #176, opened 2023-02-27
Vi/online status

### LINKED ISSUE #61: USER ACCOUNT - online status
Connect a socket to the backend
sockets are stored in a Record<user_id, Record<socket_id,socket>>
on connect, added to sockets[user_id][socket_id]
if sockets[user_id] is empty, user is offline
if there is at least 1 socket connected, user is not offline
this stores nothing in the database

OLD IDEA
-----

Use the [Page VIsibility API](https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API) to attach an event to any change to the `document.hidden` variable. 

- [ ] Add increment counter to user to count how many times use is logged in
- [ ] increment on login
- [ ] decrement on logout
- [ ] in/decrement on `document.hidden` change
- [ ] Transmit an event through websockets to change the status of the user
- [ ] user is "logged in" when counter > 0

vincentvis:
keeps track of online status through sockets
uses socket connect/disconnect to determine if someone is "online", that's all folks

closes #61


Ryno95 (approved):
Yay tabIndexes for daaaays!

I'm finally at the point where I feel I can read J/TS, man but that doesn't mean I like it more, quite the contrary 😆 

Looks good and it seems to work smoothly! 👍

livlaming (approved):

PR #180, opened 2023-03-02
Vi/2fa settings

vincentvis:
The settings page for turning on and off 2fa

refactored seed.data a bit to actually use id's..
changed router/navigation a bit to allow nested pages + ball in menu still visible


backend/src/auth/twofa/twofa.service.ts
```diff
@@ -0,0 +1,52 @@
...
+import * as QRCode from 'qrcode';
+import { authenticator } from 'otplib';
+import { TwoFaDto } from './dto/twofa.dto';
+
+// we might want to consider encrypting/decrypting the secret
+// instead of saving it plaintext into the db
+//
+// import { scrypt, randomFill, createCipheriv } from 'crypto';
+// https://nodejs.org/api/crypto.html#class-cipher
+//
```
Ryno95, line 14: I agree,  not sure if it's a little out of scope though?

Ryno95 (approved):
Biem! i feel like you are officially transcending this project ;)

backend/src/auth/twofa/twofa.service.ts
vincentvis: mjahshah, the eval sheet says to "verify that passwords are hashed"
and this is basically a password..
even though "hashing" would definitely not work here, since we need the actual value...

encrypting and decrypting it shouldn't be that much work tbh. but I would put it last on the list of things to do before Tuesday..

backend/src/auth/twofa/twofa.service.ts
```diff
@@ -0,0 +1,52 @@
...
+	async generateSecret() {
+		const secret = authenticator.generateSecret();
+		const otpauthUrl = authenticator.keyuri('userName', 'liryvianPong', secret);
+
+		const qr = await QRCode.toDataURL(otpauthUrl)
+			.then((url) => {
+				return url;
+			})
+			.catch((err) => {
+				// not sure how to handle this or how this errors..
```
livlaming, line 32: I think this line can be removed

livlaming (approved):

PR #184, opened 2023-03-02
Set DM to otherUser name

### LINKED ISSUE #185: Change DM chat title to name of other user

vincentvis:
Show username instead of chat name for DM's

closes #185


Ryno95 (approved):
Nois! Met Profile routing en de nav bar balletje bij de profiles!

PR #194, opened 2023-03-03
Ry/game with rooms bishes

### LINKED ISSUE #24: GAME - Play live Pong
Users should be able to play a live Pong game versus another player directly on the website. It can be a canvas game, or it can be a game rendered in 3D, it can also be ugly, but in any case, it must be faithful to the original Pong (1972).

TODO: Determine what will be the rules of the game, and if they are exactly the same as in the original Pong game.

**Rules original game:**
Pong is one of the first computer games that ever created. This simple "tennis like" game features two paddles and a ball. The player controls an in-game paddle by moving it vertically across the left or right side of the screen. The goal is to defeat your opponent by being the first one to gain 10 point. A player gets a point once the opponent misses a ball. The game can be played with two human players, or one player against a computer controlled paddle. 

_First refinement result:_

- [x] #38
- [x] #40
- [x] #41
- [x] #42
- [x] #70 
- [x] #69 
- [x] #68 
- [x] #71 
- [ ] #72 
- [x] #150

Ryno95:
Sooooo I've added rooms to the game. Each paddle can now only be controlled by the actually user it belongs to, which is nice.

Accidentally also added my userRelationships fix in this branch, hope  that is ok.

To play a game, check which games you have seeded, then go to /pong/:gameId with both users, this will automatically start the game.

DOES NOT WORK IN FIREFOX PRIVATE! WINDOW :)

Closes #24


frontend/src/components/pongGame/pongGame.vue
```diff
@@ -248,41 +247,56 @@ export default defineComponent({
...
+			document.addEventListener('keyup', this.keyUp);
+		}
+		this.socket.emit('joinGameRoom', this.getCurrentGame);
+		this.setSocketOn();
+		
 		// MAIN GAME LOOP
-		this.startGameLoop();
+		this.socket.on("GameCanStart", () => {
+			this.startGameLoop();
+		})
```
vincentvis, line 299: Since this should only happen once per player, you can use `this.socket.once(...`
that would create an event listener that automatically removes itself after it is called the first time

frontend/src/stores/gameSocketStore.ts
```diff
@@ -0,0 +1,26 @@
...
+	actions: {
+		async initialize() {
+			if (this.isInitialized) {
+				return;
+			}
+			this.isInitialized = true;
+			(this.socket = io('http://localhost:8080/pong', {
+				withCredentials: true,
+			})),
+				console.log('Game Socket is Initialized:', this.isInitialized);
```
vincentvis, line 23: something weird happened here with formatting..
i think there might have been a comma instead of a semicolon and prettier made it awkward

backend/src/pong/pong.gateway.ts
```diff
@@ -10,117 +11,142 @@ import {
...
-			client.join(roomName);
-			this.gameState.roomName = roomName;
-			if (this.playerOneIsInGame && this.playerTwoIsInGame) {
-				this.server.in(roomName).emit("GameCanStart");
+			if (game.id && game.player_one && game.player_two) {
+				if (!gameSubscribers[game.id]) {
+					this.createGameInstance(game, client);
+				} else {
+					client.join(gameSubscribers[game.id].roomName);
+				}
```
vincentvis, line 70: i think it's neater to not join in the `createGameInstance` method (since it's not really part of creating the instance)
and instead do this:
```ts
if (!gameSubscribers[game.id]) {
    this.createGameInstance(game, client);
}
client.join(gameSubscribers[game.id].roomName);
```

basically make sure that the instance exist and always join afterwards

backend/src/pong/pong.gateway.ts
```diff
@@ -10,117 +11,142 @@ import {
...
 			}
 		} catch (e) {}
 	}
 
-	sendPositionOfElements(@MessageBody() gameState: GameState) {
-		this.server.in(this.gameState.roomName).emit('elementPositions', gameState);
+	sendPositionOfElements(gameState: GameState) {
+		this.server
+			.in(gameSubscribers[gameState.gameId].roomName)
```
vincentvis, line 95: should this have a check if gameSubscribers[gameState.gameId] exists?
not sure if it's possible to get here after the room has been destroyed...
also not sure if `this.server.in("non_existant_room").emit()` throws or just does nothing, because if it just silently fails it's not a problem at all

vincentvis: oh no.. never mind
than we still would have to protect accessing the `roomName` property of something undefined...

Ryno95: In the "controller" on line 124, i check if the room exists, and only then proceed to call the methods that look "unprotected"

vincentvis (approved):

PR #196, opened 2023-03-03
Vi/dm create if not exist

### LINKED ISSUE #193: Create new DM chat when navigate from profilelist to non-existing dm converstation

### LINKED ISSUE #195: Create "channel owner" permission, channel owner can not be removed from channel
options:

1. add user relation to chat "owner"
2. add permission "owner" which is assigned when you create a channel

vincentvis:
Added:
- owner permission in enum
- owner permission in seed data
- logic to make sure only 1 DM between 2 users exists
- chatlist creation emit messages to subset of users or all depending on needs
- loading view for when a chat is created (with a manual delay to make it more likely that the socket is initialized..)

Refactored a few things
- chatlist routerlink
- how chat url is build
  - the dm's now use the ID of the other user in the URL instead of the ID of the chat (since the DM can possibly not yet exist)  
  - the channels still use the chat.id
- `Avatar` from `ProfilesList` because I was working on doing routerlink stuff

closes #195
closes #193


frontend/src/App.vue
```diff
@@ -8,7 +8,8 @@ import { useSocketStore } from './stores/socketStore';
 
 if (useUserStore().isLoggedIn) {
 	useUserStore().refreshMe();
-	useSocketStore().initializeOnline();
+	const socketStore = useSocketStore();
+	socketStore.initialize();
```
Ryno95, line 12: Why did you declare a socketstore first? Does this have any effect on availability of the data? (Cause it's still a big frustration with the game hihi)

frontend/src/stores/userStore.ts
```diff
@@ -105,7 +105,7 @@ export const useUserStore = defineStore('users', {
 			try {
 				await getRequest('logout');
 				this.isLoggedIn = false;
-				useSocketStore().deinitializeOnline();
+				useSocketStore().disconnect();
```
Ryno95, line 108: 👍

Ryno95 (approved):
Very nice! Feels like you've ound your flow?
 1 Curiosity comment

frontend/src/App.vue
vincentvis: that was the hope, but I think it didn't do anything
so I also call initialize (if needed) before creating the DM..

the actual "solve" was that i added a delay before posting the new channel from the frontend... I have it 1 extra second, just to make sure the socket store had some time to get started... not sure if that would also help in your case...

PR #197, opened 2023-03-03
Li/new channel

livlaming:
Alright new-channel is done!

not sure what to say about it besides, that you can create a channel via the plus sign in the side-bar of the chat page


frontend/src/stores/chatStore.ts
```diff
@@ -37,6 +52,55 @@ export const useChatStore = defineStore('chats', {
...
+					(chat: Chat_List_Item) => chat.name === createNewChannelForm.name,
+				);
+				if (found !== undefined){
+					this.errors.push('channel name already exists');
+					return;
+				}
+				if (!createNewChannelForm.name ){
+					this.errors.push('Not a valid channel name');
+					return;
+				}
```
vincentvis, line 69: can you add `createNewChannelForm.name.trim().length === 0`?
For when someone adds a name with only spaces...

frontend/src/stores/chatStore.ts
```diff
@@ -37,6 +52,55 @@ export const useChatStore = defineStore('chats', {
...
+				if (!createNewChannelForm.users[0]){
+					this.errors.push('you need to assign users to this channel');
+					return;
+				}
```
vincentvis, line 73: You don't really, right? Technically you yourself will always be in it (as the owner), so if you just want to create a channel with just you, you could...
but 🤷 maybe it's nice to have this message anyway ;P

frontend/src/views/NewChannelView.vue
```diff
@@ -1,7 +1,128 @@
...
 	<div class="page_box_wrapper">
 		<div class="page_box">
-			<h1>This is a new channel page</h1>
+			<h1>NEW CHANNEL</h1>
+			<div class="c_block c_form_group">
+				<form
+					method="Post"
+					action=""
+					class="c_block c_form_group"
+					@submit.prevent="createNewChannel(createNewChannelForm)"
```
vincentvis, line 10: 2 nested elements with the same classes, won't do much in this specific case
you can remove the div surrounding the form

frontend/src/stores/chatStore.ts
```diff
@@ -37,6 +52,55 @@ export const useChatStore = defineStore('chats', {
 			}
 		},
 
+		async createNewChannel(createNewChannelForm: CreateNewChannelForm) {
```
vincentvis, line 55: Sooooo i was thinking
this is not really functionality of the chatStore 
it just calls an external api to save something, and that api updates the chat store through sockets
so I though it is better fitted in the component itself

I made a branch `Vi/new_channel_suggestion`
didn't change any code, just moved some parts where I think they make more sense, as a suggestion
I feel medium strongly about this, but if you have a good argument why it should be in the store I'm probably easy to convince :P 
If it turns out that the suggestion on my branch messes anything up, we'll just go for it in the store it shouldn't become extra work!

frontend/src/views/NewChannelView.vue
```diff
@@ -1,7 +1,128 @@
...
+						<template v-for="user in usersWithoutMe">
+							<div class="checkbox_overview">
+								<input
+									v-model="createNewChannelForm.users"
+									type="checkbox"
+									:id="user.id"
+									name="users"
+									:value="user.id"
+								/>
+								<label :for="user.id">{{ user.name }}</label>
```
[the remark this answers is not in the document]
Ryno95: Samesies

Ryno95 (approved):
Nice I can see your new "flow"! To infinity and beyond! :D

frontend/src/stores/chatStore.ts
livlaming: like this?

if (found !== undefined && createNewChannelForm.name.trim().length === 0)

frontend/src/stores/chatStore.ts
```diff
@@ -37,6 +52,55 @@ export const useChatStore = defineStore('chats', {
...
+						permissionsEnum.OWNER,
+					]
+				})
+				const newChannel = await postRequest('chats', {
+					...createNewChannelForm,
+					users: usersToSend
+				});
+				router.push({
+					name: 'singlechat',
+					params: { currentChat : newChannel.data.id },
```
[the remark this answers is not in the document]
livlaming: no this is vincents naming... haha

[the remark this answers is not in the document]
livlaming: but maybe he did changed it by now..

frontend/src/stores/chatStore.ts
livlaming: Ha i totally agree at first I wanted to work with the chat.service etc to do some checks but I ended up doing it like this because it felt quick and easy, I'll look into your suggestions, for now maybe merge this and then try yours if it doenst fuck things up, merge yours?

frontend/src/stores/chatStore.ts
livlaming: I agree! but if I take this out it looks like something goes wrong in showing the channel?
keep this in? if you know a quick solution I'm fine with it.
I did change the message that the channel needs adding one or more members so the error is more accurate?
but maybe you don't agree with the conditions..

PR #199, opened 2023-03-03
Vi/filter blocked messages

### LINKED ISSUE #188: Filter out messages from blocked users in channels

vincentvis:
Relatively poorly tested, but should not show messages of blocked users anywhere

closes #188


frontend/src/stores/socketStore.ts
```diff
@@ -41,7 +41,6 @@ export const useSocketStore = defineStore('sockets', {
 					useChatStore().socketAction(update);
 				},
 			);
-			this.chats.socket.on('');
```
Ryno95, line 44: 👍

Ryno95 (commented):
I'm assuming private messages don.t count because that chat wont show up in your list anymore? (Cause there it doesn't work i'll elaborate if needed)

The rest looks good!

livlaming (approved):

PR #200, opened 2023-03-04
Vi/channel password

### LINKED ISSUE #186: Channel password protection

vincentvis:
channels can be password protected now
re-seed database to get a chat with a password

does not actually protect the messages endpoint.. which is kinda "bad".. should probably, if we do this correctly, have the password required in a parameter when you request the messages from the chat that is password protected.. but that is more work.. :P 

closes #186


Ryno95 (commented):
It looks like the password is only needed by one user, afterwards anyone can join freely. Is what I thought, but then I saw I was in the main branch looking at the intra login hahaha

Looks good! Maybe I can look at it again once the other branches have been merged? So I can zoom in on the specific files for this branch?

livlaming (approved):
I checked this one by just runing the thing, so we can move forward quickly, password and error seem to worken perfectly!

PR #201, opened 2023-03-04
Ry/active games

### LINKED ISSUE #28: GAME - Watch games
The user should be able to watch a live play between other users without interfering with it.

Ryno95:
The default game page now show a list of active games.

Clicking on watch game will later route you to a specific game as a watcher.

CSS courtesy of Señor Vincentos ;)

Closes #28


frontend/src/components/pongGame/ActiveGameRow.vue
```diff
@@ -0,0 +1,30 @@
+<template>
```
livlaming, line 1: the idee of the element/component OverViewWithMidline is that you can use component from a 'page component' and 'feed it a data array with left and right.

<OverviewWithMidline :data-array="dataArray" />

this.dataArray = [
				{ left: 'intra name', right: this.user.name },
				{ left: 'member since', right: this.user.created_at },
				{ left: 'wins', right: 1 },
				{ left: 'losses', right: 190 },

you can look at profileview for an example

frontend/src/assets/main.css
```diff
@@ -569,3 +569,53 @@ h1 {
 .grayedOut:hover {
 	text-decoration: none;
 }
+
```
livlaming, line 572: I don't think these should be in the main.css because it's connected to this specific component.
At least I wrote it like this.

@vvissche probably knows best what the standerd way of working is

Ryno95: This is used in quite a few components.

vincentvis (approved):

PR #207, opened 2023-03-06
Vi/game request message

vincentvis:
makes the message that is send clickable

### DIFF
--- a/frontend/src/components/chat/Message.vue
+++ b/frontend/src/components/chat/Message.vue
@@ -1,7 +1,10 @@
 <template>
 	<div class="c_message" :class="{ 'c_message--mine': isMessageMine }">
 		<div class="c_message__name">{{ getName }}</div>
-		<div class="c_message__wrap">
+		<div
+			class="c_message__wrap"
+			:class="{ is_game_request: isGameRequest }"
+		>
 			<div class="c_message__msg">
 				<div class="c_message__time">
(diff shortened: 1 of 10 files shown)


frontend/src/components/chat/Message.vue
```diff
@@ -38,6 +49,12 @@ export default defineComponent({
 		};
```
livlaming, line 49: I cant comment on line 45 but it says:

45: // shouldbe removed in favour of sockets + init once 

is this something we can do now?

frontend/src/views/RequestGame.vue
```diff
@@ -89,16 +89,58 @@ export default defineComponent({
...
-			errors,
-			createGame,
 			allUsers,
 			createGameForm,
 			newMessage,
 		};
 	},
+	data() {
+		return {
+			errors: [] as String[],
```
livlaming, line 104: just a question, why did you moved the errors? *if this takes to much time to reply on e can discuss this later!

livlaming (commented):
I see you moved the creategame() from gamestore to the gameRequest, 
Currently I am using the createGame for the match making. I can rewrite this surely, but what is the advantage of moving this outside the store?
And would you like me to repeat this in matchmaking or does it makes sense to make in generic?

livlaming (approved):

PR #208, opened 2023-03-06
Wins and losses now reflect real values

### LINKED ISSUE #202: update/get game info for ProfileView

Ryno95:
Wins and losses now reflect real values.

New endpoint in user.controller, thought of doing it as a generated column, buuuuut had issues with id's of users, also realized that it doesn't need to be calculated the entire time, only when needed

closes #202


livlaming (approved):
Haha data base playeeerrrrr, 
it's hard to check since the game isn't there and for some reason my data base is hiding the wins and losses, BUT I trust you and I'll double check it lateerrrrrrr.

PR #224, opened 2023-03-07
fix 2fa

vincentvis:
fixes 2fa flow

### DIFF
--- a/frontend/src/views/LoginView.vue
+++ b/frontend/src/views/LoginView.vue
@@ -64,7 +64,6 @@ export default defineComponent({
 			try {
 				this.loginStage = 'loading';
 				const user: User = (await postRequest('login', loginForm)).data;
-				console.log({ user });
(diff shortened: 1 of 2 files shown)


frontend/src/views/LoginView.vue
```diff
@@ -64,6 +64,7 @@ export default defineComponent({
 			try {
 				this.loginStage = 'loading';
 				const user: User = (await postRequest('login', loginForm)).data;
+				console.log({ user });
```
Ryno95, line 67: rm

Ryno95 (commented):
my 1st github mobile review🤓

Ryno95 (approved):

frontend/src/views/LoginView.vue
vincentvis:
```suggestion
```

